PR #10, opened 2020-01-21
Feature/list mini apps

ts-leojoseph-r:
PR for fetching the list of Mini Apps and BDD tests


MiniApp/Classes/MiniAppLister.swift
```diff
@@ -1,2 +1,54 @@
...
+    var environment: Environment
+    var miniAppClient: MiniAppClient
+
+    init(environment: Environment) {
+        self.environment = environment
+        self.miniAppClient = MiniAppClient()
+    }
+
+    func fetchList(completionHandler: @escaping (Result<[MiniAppInfo], Error>) -> Void) {
+        guard let url = environment.listingUrl else {
```
corycaywood: I'm not sure that `Environment` is the right place for `listingUrl`. The `baseUrl` is part of the environment, but the listing URL is part of the API request, so it should closer to the API part of the code.

MiniApp/Classes/MiniAppLister.swift
```diff
@@ -1,2 +1,54 @@
...
+                    return completionHandler(.failure(self.invalidResponseData()))
+                }
+                return completionHandler(.success(decodeResponse))
+            case .failure(let error):
+                return completionHandler(.failure(error))
+            }
+        }
+    }
+
+    func listingURLRequest(url: URL) -> URLRequest {
```
corycaywood: I kind of would like to see more separation between the API request code and the business logic parts of the code. 

Right now, this isn't a big problem because the business logic for getting the mini app list is very simple. But with something more complicated like `MiniAppDownloader`, there will be a lot more business logic so I don't think we should be mixing the API code with it. Also, there could be more complicated logic required in the future for `MiniAppLister`.

On Android, all of the API code is in a separate `api` package: https://github.com/rakutentech/android-miniapp/tree/master/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api
And then `MiniAppLister` is very simple: https://github.com/rakutentech/android-miniapp/blob/master/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppLister.kt

So maybe most of the code that's currently in `MiniAppLister` could be moved to a separate `ListRequest` class and then `MiniAppLister` can just handle converting the response to a list of `MiniAppInfo`? What do you think?

MiniApp/Classes/MiniAppLister.swift
```diff
@@ -1,2 +1,54 @@
...
+
+    func fetchList(completionHandler: @escaping (Result<[MiniAppInfo], Error>) -> Void) {
+        guard let url = environment.listingUrl else {
+            return completionHandler(.failure(self.invalidURLError()))
+        }
+
+        self.miniAppClient.requestFromServer(request: listingURLRequest(url: url)) { (result) in
+            switch result {
+            case .success(let responseData):
+                guard let decodeResponse = self.decodeListingResponse(with: responseData.data) as? [MiniAppInfo] else {
```
corycaywood: I think there needs to be a separate model for the response rather than decoding directly to `MiniAppInfo`, and then you can convert the response model to `MiniAppInfo`. This is so we can avoid having the public model be dependent on the parameter names used by the API and avoid exposing things from the API response which shouldn't be exposed to the user. 

On Android, we were able to use a single class for both and just change some of the parameters to `internal`, and we can use `SerializedName` if we wish to have different parameter names than the response. Is something similar possible on iOS? 
https://github.com/rakutentech/android-miniapp/blob/master/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppInfo.kt

Also, the response model from the API has changed, so please check it again.

Tests/Unit/MiniAppListerTests.swift
```diff
@@ -0,0 +1,113 @@
...
+            }
+            context("when request from server returns invalid data") {
+                it("will decode the response with MiniAppInfo decodable") {
+                    let mockAPIClient = MockAPIClient()
+                    var testError: NSError?
+                    mockBundle.mockEndpoint = "http://www.example.com"
+                    let miniAppLister = MiniAppLister(environment: Environment(bundle: mockBundle))
+                    miniAppLister.miniAppClient = mockAPIClient
+                    let responseString = """
+                    [{"test": "123", "name": "Test", "description": "Test", "icon": "https://test.com",
```
corycaywood: Can you make this formatted JSON so each key/value is on it's own line?

MiniApp/Classes/MiniAppLister.swift
```diff
@@ -1,2 +1,54 @@
...
+                return completionHandler(.success(decodeResponse))
+            case .failure(let error):
+                return completionHandler(.failure(error))
+            }
+        }
+    }
+
+    func listingURLRequest(url: URL) -> URLRequest {
+        var urlRequest = URLRequest(url: url)
+        urlRequest.setAuthorizationHeader(environment: environment)
```
corycaywood: Is it possible to set this in the `MiniAppClient` so you don't have to add it to every API request?

ts-leojoseph-r: We are preparing the URLRequest in MiniAppLister with listingURL, Headers and Query parameters. Moving just this wouldn't be odd ?
Even moving to MiniAppClient will set the authorization for every API request.

Tests/Unit/MiniAppListerTests.swift
ts-leojoseph-r: Done

MiniApp/Classes/MiniAppLister.swift
```diff
@@ -1,2 +1,54 @@
 internal class MiniAppLister {
+    var environment: Environment
+    var miniAppClient: MiniAppClient
+
+    init(environment: Environment) {
+        self.environment = environment
+        self.miniAppClient = MiniAppClient()
+    }
+
+    func fetchList(completionHandler: @escaping (Result<[MiniAppInfo], Error>) -> Void) {
```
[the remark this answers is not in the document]
donnie-jp: Even in Swift a completion closure is still AFAIK the best option for async. There's no suspend  equivalent that I know of. You could still force the app to handle concurrency by implicitly blocking but I don't think we'd want to do that.

MiniApp/Classes/Extensions/URLRequest+Authorization.swift
```diff
@@ -0,0 +1,7 @@
+extension URLRequest {
+    mutating func setAuthorizationHeader(environment: Environment) {
+        if environment.subscriptionKey.count > 0 {
```
donnie-jp: Using https://developer.apple.com/documentation/swift/string/2946268-isempty would be better

MiniApp/Classes/MiniAppClient.swift
```diff
@@ -1,50 +1,66 @@
...
-        }
+protocol MiniAppClientProtocol {
+    func startDataTask(with request: URLRequest, completionHandler: @escaping (Data?, URLResponse?, Error?) -> Void)
+}
 
-        let (responseBody, metadata) = serviceCommunicator.requestFromServer(withUrl: manifestUrl,
-                                                                             withHttpMethod: .get,
-                                                                             withSemaphoreWait: true)
+extension URLSession: MiniAppClientProtocol {
+    func startDataTask(with request: URLRequest, completionHandler: @escaping (Data?, URLResponse?, Error?) -> Void) {
```
donnie-jp: Can you use a `Result` type for the completion handler and handle the optionals inside the function?

MiniApp/Classes/MiniAppClient.swift
```diff
@@ -1,50 +1,66 @@
...
+}
 
-    func fetchManifest(with appId: String, and versionId: String) -> ManifestResponse? {
-        guard let manifestUrl = formRequestUrl(with: appId, and: versionId) else {
-            #if DEBUG
-                print("MiniAppSDK: Failed to create manifest URL.")
-            #endif
-            return nil
-        }
+protocol MiniAppClientProtocol {
```
donnie-jp: Naming seems weird. `SessionProtocol` ?

MiniApp/Classes/MiniAppLister.swift
donnie-jp: agree with Cory

MiniApp/Classes/MiniAppLister.swift
donnie-jp: I think you can set it on the session https://developer.apple.com/documentation/foundation/urlsessionconfiguration/1411532-httpadditionalheaders

MiniApp/Classes/MiniAppLister.swift
```diff
@@ -1,2 +1,54 @@
...
+
+    func listingURLRequest(url: URL) -> URLRequest {
+        var urlRequest = URLRequest(url: url)
+        urlRequest.setAuthorizationHeader(environment: environment)
+        return urlRequest
+    }
+
+    func decodeListingResponse(with dataResponse: Data?) -> [MiniAppInfo]? {
+        do {
+            return try JSONDecoder().decode(Array<MiniAppInfo>.self, from: dataResponse!) as [MiniAppInfo]
```
donnie-jp: Don't force unwrap. Use a guard at the top of the function.

MiniApp/Classes/Models/MiniAppInfo.swift
```diff
@@ -1,7 +1,13 @@
...
+public struct MiniAppInfo: Decodable {
+    public var id: String
+    public var name: String
+    public var description: String
+    public var icon: URL
+    public var version: Version
+}
+
+public struct Version: Decodable {
+    public var versionTag: String
```
corycaywood: Ah, I didn't realize that the spec for this changed again. We will need to update this on Android...

MiniApp/Classes/Extensions/URLRequest+Authorization.swift
ts-leojoseph-r: Done

MiniApp/Classes/MiniAppLister.swift
ts-leojoseph-r: Done

MiniApp/Classes/MiniAppLister.swift
ts-leojoseph-r: Done.

MiniApp/Classes/MiniAppLister.swift
ts-leojoseph-r: I think that's a good approach to have them separated like API request and business logic. 
I was thinking to create some separate class to handle API request. Will work on this new approach in new PR

MiniApp/Classes/MiniAppLister.swift
ts-leojoseph-r: We should not be modifying the headers in URLSession object as per the documentation.
"An URLSession object is designed to handle various aspects of the HTTP protocol for you. As a result, you should not modify the following headers:
* Authorization"

MiniApp/Classes/MiniAppClient.swift
ts-leojoseph-r: Yes we can. I was thinking the same to have Result type, but as of now dataTask methods of URLSession doesn't support any result types and by doing above changes I might end up doing most of the error handling in the extension itself (which is good though). But I thought I should follow similar standards like Remote config.

MiniApp/Classes/MiniAppClient.swift

PR #11, opened 2020-01-23
Sample application to show list of mini apps

ts-leojoseph-r:
Sample application with table view to show list of mini apps. Tableview displays the following items,
* Mini app icon
* Mini app title
* Mini app description


Example/Images.xcassets/image_placeholder.imageset/Contents.json
```diff
@@ -0,0 +1,21 @@
+{
+  "images" : [
+    {
+      "idiom" : "universal",
+      "filename" : "2.png",
```
corycaywood: This is a placeholder image that will be shown before the icon has finished downloading?

Example/Info.plist
```diff
@@ -2,6 +2,11 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>NSAppTransportSecurity</key>
+	<dict>
+		<key>NSAllowsArbitraryLoads</key>
```
corycaywood: Setting this will allow `HTTP` connections? Is there a reason this is needed?

Example/ViewController.swift
```diff
@@ -1,9 +1,88 @@
...
+                self.tableView.reloadData()
+                self.dismiss(animated: false, completion: nil)
+            case .failure(let error):
+                print(error.localizedDescription)
+                self.displayAlert(title: "Error", message: "Couldn't retrieve Mini App list, please try again later")
+            }
+        }
+    }
+
+    func decodeListingResponse(with dataResponse: Data?) -> [MiniAppInfo]? {
```
corycaywood: I don't see anyplace where this function is used. Seems like something that shouldn't be needed?

Example/Images.xcassets/image_placeholder.imageset/Contents.json
ts-leojoseph-r: Yes, Looks like the image name inside the json was wrong. I have updated in the new commit now.

Example/ViewController.swift
ts-leojoseph-r: Yes. Removed them now

Example/Info.plist
ts-leojoseph-r: No. It will allow only when NSExceptionAllowsInsecureHTTPLoads is added in NSAppTransportSecurity

donnie-jp: Why do we need it? The backend doesn't have proper https / cert?

@corycaywood this is a workaround that allows all connections, including insecure ones.

ts-leojoseph-r: I thought the staging is not gonna have one. Looks it is having valid one. Will update the plist.

MiniApp.xcodeproj/project.pbxproj
```diff
@@ -17,6 +17,8 @@
 		607FACDB1AFB9204008FA782 /* Main.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 607FACD91AFB9204008FA782 /* Main.storyboard */; };
 		607FACE01AFB9204008FA782 /* LaunchScreen.xib in Resources */ = {isa = PBXBuildFile; fileRef = 607FACDE1AFB9204008FA782 /* LaunchScreen.xib */; };
 		607FACEC1AFB9204008FA782 /* Tests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 607FACEB1AFB9204008FA782 /* Tests.swift */; };
+		C3866B58983DE5A1C9FC45CF /* Pods_MiniApp_Tests.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 6E1CC4976FCCFA9C5737280C /* Pods_MiniApp_Tests.framework */; };
+		F8673656152B3624D9E6EAAB /* Pods_MiniApp_Example.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = FB348493F8CB435AB6C5F86C /* Pods_MiniApp_Example.framework */; };
```
donnie-jp: We don't usually commit with Pods integrated but it's up to you if this is easier. There are pros/cons to both approaches.

ts-leojoseph-r: Sure. Will consider this in the next commit.

PR #15, opened 2020-01-28
Fetching manifest details of mini app

ts-leojoseph-r:
Updated fetching manifest details of mini app and respective test cases

### DIFF
--- a/MiniApp/Classes/MiniAppClient.swift
+++ b/MiniApp/Classes/MiniAppClient.swift
@@ -8,12 +8,14 @@ protocol SessionProtocol {
 class MiniAppClient {
     let session: SessionProtocol
     let listingApi: ListingApi
+    let manifestApi: ManifestApi
     let environment: Environment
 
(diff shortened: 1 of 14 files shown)


MiniApp/Classes/MiniAppClient.swift
```diff
@@ -26,17 +28,52 @@ class MiniAppClient {
...
+                        completionHandler: @escaping (Result<ManifestResponse, Error>) -> Void) {
+
+        guard let urlRequest = self.manifestApi.createURLRequest(appId: appId, versionId: versionId) else {
+            return completionHandler(.failure(NSError.invalidURLError()))
+        }
+
+        session.startDataTask(with: urlRequest) { (result) in
+            switch result {
+            case .success(let responseData):
+                if !(200...299).contains(responseData.httpResponse.statusCode) {
```
corycaywood: It looks like this `if` statement and its contents are the same in `getMiniAppsList`. Is there any way to reduce the code duplication? Maybe everything in the `.success` case could be pulled out into some kind of generic `handleSuccessResponse` function?

PR #16, opened 2020-02-03
Download API request changes

ts-leojoseph-r:
PR for creating API for starting the download API in client.


MiniApp/Classes/Extensions/MiniApp+URLRequest.swift
```diff
@@ -0,0 +1,9 @@
+extension URLRequest {
+    static func createURLRequest(url: URL, environment: Environment) -> URLRequest {
+        var urlRequest = URLRequest(url: url)
+        if environment.subscriptionKey.isEmpty {
```
corycaywood: Isn't this backwards? It should be setting the subscription key if it's not empty?

MiniApp/Classes/Extensions/MiniApp+URLRequest.swift
```diff
@@ -0,0 +1,9 @@
+extension URLRequest {
+    static func createURLRequest(url: URL, environment: Environment) -> URLRequest {
```
corycaywood: Are there any unit tests for this?

MiniApp/Classes/Extensions/MiniApp+URLRequest.swift
ts-leojoseph-r: Yes, you were right. Updated the code now.

MiniApp/Classes/Extensions/MiniApp+URLRequest.swift
ts-leojoseph-r: Xcode was showing that the code coverage as 100% for this extension but there was no dedicated testcase written for this file. Surprisingly testcases were passed too. Now I have created a Unit testcases for createURLRequest

donnie-jp: It can be misleading to rely on the coverage percentage to determine if a function/class/whatever has got specific tests to verify its behaviour.

ts-leojoseph-r: Noted

MiniApp/Classes/MiniAppClient.swift
```diff
@@ -3,21 +3,27 @@ protocol SessionProtocol {
...
-        self.session = session
+    override init() {
         self.environment = Environment()
         self.listingApi = ListingApi(environment: self.environment)
         self.manifestApi = ManifestApi(environment: self.environment)
+        self.downloadApi = DownloadApi(environment: self.environment)
     }
 
+    lazy var session: SessionProtocol = {
+        return URLSession(configuration: .default, delegate: self, delegateQueue: nil)
```
donnie-jp: I don't think it should _always_ use the default session configuration. The session should probably be injected as a dependency.

MiniApp/Classes/MiniAppClient.swift
```diff
@@ -3,21 +3,27 @@ protocol SessionProtocol {
...
+    let downloadApi: DownloadApi
     let environment: Environment
 
-    init(session: SessionProtocol = URLSession(configuration: URLSessionConfiguration.default)) {
-        self.session = session
+    override init() {
         self.environment = Environment()
         self.listingApi = ListingApi(environment: self.environment)
         self.manifestApi = ManifestApi(environment: self.environment)
+        self.downloadApi = DownloadApi(environment: self.environment)
```
donnie-jp: Is this where you're doing dependency setup? Seems odd to do it in the client.

MiniApp/Classes/MiniAppClient.swift
ts-leojoseph-r: We had that logic earlier by creating URLSession with default configuration in the constructor. 

`init(session: SessionProtocol = URLSession(configuration: URLSessionConfiguration.default)) `

But now we need to create URLSession object that handles delegates for downloading data task which wouldn't allow to create like below,

`init(session: SessionProtocol = {
        return URLSession(configuration: .default, delegate: self, delegateQueue: nil)
    }()) {`

That's why I was using lazy variable of URLSession. Should I use different APIClient objects for downloading and other?

MiniApp/Classes/MiniAppClient.swift
ts-leojoseph-r: These classes were meant to create appropriate URL request for listing, manifest and downloading, which will be accessed only by the APIClient. Initially these changes were in Environment class, but later it has been moved to different classes to follow the architectural implementation of mini app

MiniApp/Classes/MiniAppClient.swift
donnie-jp: ok, thanks, then I think your way is ok - but maybe revisit this later to see if there's a better approach

ts-leojoseph-r: Yeah sure. May be this could change after MVP once we bring in State machine for downloading.

PR #17, opened 2020-02-05
Feature/downloading mini app

ts-leojoseph-r:
Base implementation of downloading mini app files and saving them.

### DIFF
--- a/MiniApp/Classes/Extensions/MiniApp+NSError.swift
+++ b/MiniApp/Classes/Extensions/MiniApp+NSError.swift
@@ -35,7 +35,7 @@ extension NSError {
         return NSError(
             domain: "Downloader",
             code: 0,
-            userInfo: [NSLocalizedDescriptionKey: "Mini App Downloading failed"]
+            userInfo: [NSLocalizedDescriptionKey: "Downloading failed"]
         )
     }
 }
--- a/MiniApp/Classes/Extensions/MiniApp+NSError.swift
+++ b/MiniApp/Classes/Extensions/MiniApp+NSError.swift
@@ -30,4 +30,12 @@ extension NSError {
             userInfo: [NSLocalizedDescriptionKey: "Invalid response received"]
         )
     }
+
+    class func downloadingFailed() -> NSError {
+        return NSError(
+            domain: "Downloader",
+            code: 0,
+            userInfo: [NSLocalizedDescriptionKey: "Mini App Downloading failed"]
+        )
+    }
 }
--- a/MiniApp/Classes/MiniAppClient.swift
+++ b/MiniApp/Classes/MiniAppClient.swift
@@ -12,6 +12,7 @@ class MiniAppClient: NSObject, URLSessionDownloadDelegate {
     let manifestApi: ManifestApi
     let downloadApi: DownloadApi
     let environment: Environment
+    weak var delegate: MiniAppDownloaderProtocol?
 
     override init() {
         self.environment = Environment()
@@ -75,8 +76,11 @@ class MiniAppClient: NSObject, URLSessionDownloadDelegate {
         return NSError.serverError(code: errorModel.code, message: errorModel.message)
     }
 
-    // Saving the file - Storage
     func urlSession(_ session: URLSession, downloadTask: URLSessionDownloadTask, didFinishDownloadingTo location: URL) {
+        delegate?.downloadedFileData(downloadTask: downloadTask, location: location)
+    }
 
+    func urlSession(_ session: URLSession, task: URLSessionTask, didCompleteWithError error: Error?) {
+        delegate?.downloadCompleted(task: task, error: error)
     }
 }
--- a/MiniApp/Classes/MiniAppClient.swift
+++ b/MiniApp/Classes/MiniAppClient.swift
@@ -92,4 +92,3 @@ class MiniAppClient: NSObject, URLSessionDownloadDelegate {
         delegate?.downloadCompleted(url: url, error: error)
     }
 }
-
--- a/MiniApp/Classes/MiniAppClient.swift
+++ b/MiniApp/Classes/MiniAppClient.swift
@@ -77,10 +77,19 @@ class MiniAppClient: NSObject, URLSessionDownloadDelegate {
     }
 
     func urlSession(_ session: URLSession, downloadTask: URLSessionDownloadTask, didFinishDownloadingTo location: URL) {
-        delegate?.downloadedFileData(downloadTask: downloadTask, location: location)
+        guard let destinationURL = downloadTask.currentRequest?.url?.absoluteString else {
+            delegate?.downloadCompleted(url: "", error: NSError.downloadingFailed())
+            return
+        }
+        delegate?.fileDownloaded(sourcePath: location, destinationPath: destinationURL)
     }
 
     func urlSession(_ session: URLSession, task: URLSessionTask, didCompleteWithError error: Error?) {
-        delegate?.downloadCompleted(task: task, error: error)
+        guard let url = task.currentRequest?.url?.absoluteString else {
+            delegate?.downloadCompleted(url: "", error: NSError.downloadingFailed())
+            return
+        }
+        delegate?.downloadCompleted(url: url, error: error)
     }
 }
+
--- a/MiniApp/Classes/MiniAppDownloader.swift
+++ b/MiniApp/Classes/MiniAppDownloader.swift
@@ -1,16 +1,12 @@
-/**
- * Protocol for the Downloader class that handles downloading the files for Mini App.
- */
-protocol MiniAppDownloaderProtocol {
+class MiniAppDownloader {
 
-    /// Handles downloading each file to a specific location.
-    /// - Parameters:
-    ///   - urls: [String] - List of URLs in the manifest.
-    ///   - pathDirectory: URL - The system path directory to cache.
-    func download(_ urls: [String], to pathDirectory: URL)
-}
+    typealias DownloadCompletionHandler = (Result<URL?, Error>) -> Void
 
-class MiniAppDownloader: NSObject, MiniAppDownloaderProtocol {
+    var miniAppClient: MiniAppClient?
+    var miniAppStorage: MiniAppStorage
+    var manifestDownloader: ManifestDownloader
+    var completionHandler: DownloadCompletionHandler?
+    var miniAppDirectory: URL?
 
     private var urlToDirectoryMap = [String: URL]()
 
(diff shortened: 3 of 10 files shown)


Tests/Unit/MiniAppDownloaderTests.swift
```diff
@@ -0,0 +1,91 @@
...
+            context("when valid manifest information is returned ") {
+                it("will be downloaded and path is returned") {
+                    let mockAPIClient = MockAPIClient()
+                    let downloader = MiniAppDownloader()
+                    let responseString = """
+                      {
+                        "id": "123",
+                        "versionTag": "1",
+                        "name": "Test",
+                        "files": ["https://devstreaming-cdn.apple.com/videos/tutorials/TestFlight_App_Store_Connect_2018/TestFlight_App_Store_Connect_2018_sd.mp4",
```
corycaywood: I think that `MiniAppClient#download` isn't mocked, right? So this is downloading the real file? 

Can you add a mock for `download` and have it return a mock file or some static file from the repo?

MiniApp/Classes/Extensions/MiniApp+NSError.swift
```diff
@@ -30,4 +30,12 @@ extension NSError {
             userInfo: [NSLocalizedDescriptionKey: "Invalid response received"]
         )
     }
+
+    class func downloadingFailed() -> NSError {
```
donnie-jp: Does Android use this naming? I think `downloadFailed` and same naming in error message would be better.

MiniApp/Classes/MiniAppClient.swift
```diff
@@ -75,8 +76,11 @@ class MiniAppClient: NSObject, URLSessionDownloadDelegate {
         return NSError.serverError(code: errorModel.code, message: errorModel.message)
     }
 
-    // Saving the file - Storage
     func urlSession(_ session: URLSession, downloadTask: URLSessionDownloadTask, didFinishDownloadingTo location: URL) {
+        delegate?.downloadedFileData(downloadTask: downloadTask, location: location)
```
donnie-jp: `fileDownloaded` ?

MiniApp/Classes/MiniAppDownloader.swift
```diff
@@ -1,16 +1,12 @@
...
-    /// Handles downloading each file to a specific location.
-    /// - Parameters:
-    ///   - urls: [String] - List of URLs in the manifest.
-    ///   - pathDirectory: URL - The system path directory to cache.
-    func download(_ urls: [String], to pathDirectory: URL)
-}
+    typealias DownloadCompletionHandler = (Result<URL?, Error>) -> Void
 
-class MiniAppDownloader: NSObject, MiniAppDownloaderProtocol {
+    var miniAppClient: MiniAppClient?
```
donnie-jp: Does this need to be optional?

MiniApp/Classes/MiniAppDownloader.swift
```diff
@@ -1,16 +1,12 @@
...
+    var miniAppStorage: MiniAppStorage
+    var manifestDownloader: ManifestDownloader
+    var completionHandler: DownloadCompletionHandler?
+    var miniAppDirectory: URL?
```
donnie-jp: Why not make these dependencies private and inject them in init?

MiniApp/Classes/MiniAppDownloader.swift
```diff
@@ -22,52 +18,70 @@ class MiniAppDownloader: NSObject, MiniAppDownloaderProtocol {
...
-                #if DEBUG
-                    print("MiniAppSDK: Invalid URL from manifest.")
-                #endif
-                return
+    func download(with apiClient: MiniAppClient, appId: String, versionId: String, completionHandler: @escaping (Result<URL?, Error>) -> Void) {
+        self.miniAppClient = apiClient
+        self.manifestDownloader.fetchManifest(apiClient: apiClient, appId: appId, versionId: versionId) { (result) in
+            switch result {
+            case .success(let responseData):
+                self.completionHandler = completionHandler
```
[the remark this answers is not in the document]
donnie-jp: I think he's doing this because of the delegate handling. I'm not sure the best option here.

MiniApp/Classes/MiniAppDownloader.swift
```diff
@@ -22,52 +18,70 @@ class MiniAppDownloader: NSObject, MiniAppDownloaderProtocol {
...
 
-/**
- * URLSessionDownloadDelegate methods for MiniAppDownloader.
- */
-extension MiniAppDownloader: URLSessionDownloadDelegate {
-    func urlSession(_ session: URLSession, downloadTask: URLSessionDownloadTask, didFinishDownloadingTo location: URL) {
-        guard let sourceURL = downloadTask.originalRequest?.url else {
+extension MiniAppDownloader: MiniAppDownloaderProtocol {
+
+    func downloadedFileData(downloadTask: URLSessionDownloadTask, location: URL) {
```
donnie-jp: This seems to just need a `URL` passed in?

MiniApp/Classes/MiniAppDownloader.swift
```diff
@@ -22,52 +18,70 @@ class MiniAppDownloader: NSObject, MiniAppDownloaderProtocol {
...
         }
+        miniAppStorage.save(sourcePath: location, destinationPath: destinationPath)
+    }
 
-        try? FileManager.default.createDirectory(atPath: destinationPath.relativePath, withIntermediateDirectories: true)
-        try? FileManager.default.removeItem(at: destinationPath)
-        try? FileManager.default.copyItem(at: location, to: destinationPath)
-
-        urlToDirectoryMap.removeValue(forKey: sourceURL.absoluteString)
+    func downloadCompleted(task: URLSessionTask, error: Error?) {
```
donnie-jp: This seems to just need a URL passed in?

MiniApp/Classes/MiniAppStorage.swift
```diff
@@ -0,0 +1,7 @@
+class MiniAppStorage {
+    internal func save(sourcePath: URL, destinationPath: URL) {
+        try? FileManager.default.createDirectory(atPath: destinationPath.relativePath, withIntermediateDirectories: true)
+        try? FileManager.default.removeItem(at: destinationPath)
+        try? FileManager.default.copyItem(at: sourcePath, to: destinationPath)
```
donnie-jp: Can any of these fail? Do you need to handle/report the errors?

Tests/Unit/MiniAppDownloaderTests.swift
```diff
@@ -0,0 +1,91 @@
+import Quick
+import Nimble
+@testable import MiniApp
+
+// swiftlint:disable function_body_length
+class MiniAppDownloaderTests: QuickSpec {
+
+    override func spec() {
+        describe("mini app folder will be created") {
```
donnie-jp: Does the SDK code work if the folder already exists? Is that tested?

MiniApp/Classes/MiniAppDownloader.swift
donnie-jp: If the completionHandler references the same MiniAppDownloader instance this would cause a retain cycle.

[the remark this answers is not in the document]
ts-leojoseph-r: Yep. We need to store the completionhandler so that it can be used in the delegates whenever downloading is finished/failed

MiniApp/Classes/MiniAppDownloader.swift
```diff
@@ -22,52 +18,70 @@ class MiniAppDownloader: NSObject, MiniAppDownloaderProtocol {
...
+                self.completionHandler = completionHandler
+                self.downloadManifestFiles(with: appId, versionId: versionId, manifestResponse: responseData)
+            case .failure(let error):
+                return completionHandler(.failure(error))
             }
+        }
+    }
 
-            urlToDirectoryMap[url] = pathDirectory.appendingPathComponent(fileDirectory)
+    private func downloadManifestFiles(with appId: String, versionId: String, manifestResponse: ManifestResponse) {
```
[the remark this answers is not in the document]
ts-leojoseph-r: Done

MiniApp/Classes/MiniAppDownloader.swift
```diff
@@ -22,52 +18,70 @@ class MiniAppDownloader: NSObject, MiniAppDownloaderProtocol {
...
+        guard let miniAppStoragePath = FileManager.getMiniAppDirectory(with: appId, and: versionId) else {
+            return
+        }
+        miniAppDirectory = miniAppStoragePath
```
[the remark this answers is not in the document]
ts-leojoseph-r: Removed. We need to pass this parameter back to RealMiniApp once the download is completed and we would know the state of the download only using the delegates, that's why we are using this global variable.

Updated the logic to just return success. Once the download is completed, RealMiniApp can request the path of the MiniApp so we won't be having any issues.

Tests/Unit/MiniAppDownloaderTests.swift
ts-leojoseph-r: Updated. Mocked the download progress now by.

MiniApp/Classes/Extensions/MiniApp+NSError.swift
ts-leojoseph-r: I couldn't find anything related to this in Android side.

MiniApp/Classes/MiniAppClient.swift

MiniApp/Classes/MiniAppDownloader.swift
ts-leojoseph-r: Done

MiniApp/Classes/MiniAppDownloader.swift
ts-leojoseph-r: Done. Couldn't move everything, but moved few that can be injected. Thanks :)

MiniApp/Classes/MiniAppDownloader.swift
ts-leojoseph-r: Done

MiniApp/Classes/MiniAppDownloader.swift
ts-leojoseph-r: Done

MiniApp/Classes/MiniAppStorage.swift
ts-leojoseph-r: Updated

Tests/Unit/MiniAppDownloaderTests.swift
ts-leojoseph-r: We don't download the file if file exists already for a particular mini app.

MiniApp/Classes/MiniAppDownloader.swift
ts-leojoseph-r: Yeah, I felt the same. But I guess this can be handled by creating a state machine for downloading that could help not only to avoid these memory leaks but also to have more control over download process. This PR is for base implementation of downloading and storing the mini app. Definitely these changes will be done when we consider to update them with the state machine. Or may be we can use some 3rd party pods just for networking :wink:

MiniApp/Classes/Extensions/MiniApp+NSError.swift
corycaywood: Android uses synchronous `suspend` functions, so it doesn't have anything similar to this.

MiniApp/Classes/MiniAppDownloader.swift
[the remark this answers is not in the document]
corycaywood: I wanted to get more clarification on this this still: will you be instantiating a new `MiniAppDownloader` for each download?

If you're using a single instance of `MiniAppDownloader`, then the completion handler will change if the user starts a new download while a previous download is in progress.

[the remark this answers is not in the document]
ts-leojoseph-r: Yes, I will be instantiating a new MiniAppDownloader for each download (only for the MVP). We would be definitely needing a Downloading Manager not only to overcome this issue but also to track download progress, etc.,

ts-leojoseph-r:
Point taken that we need to handle concurrency/downloading using a manager class to avoid retain cycles and also to manage the state of the download.

PR #18, opened 2020-02-11
Feature/display mini app

ts-leojoseph-r:
This is the base implementation of the displayer feature. The changes are done to create a WKWebview that will load a URL request and the MiniAppView is returned to the hosting application.

### DIFF
--- a/MiniApp/Classes/Display/Displayer.swift
+++ b/MiniApp/Classes/Display/Displayer.swift
@@ -1,6 +1,6 @@
 internal class Displayer {
 
-    func getMiniAppView(miniAppPath: URL) -> MiniAppView {
+    func getMiniAppView(miniAppPath: URL) -> MiniAppView? {
         return RealMiniAppView.shared.getMiniAppView(miniAppPath: miniAppPath)
     }
 }
--- a/MiniApp/Classes/Display/Displayer.swift
+++ b/MiniApp/Classes/Display/Displayer.swift
@@ -1,9 +1,6 @@
 internal class Displayer {
 
-    func getMiniAppView(miniAppPath: URL?) -> MiniAppView? {
-        guard let filePath = miniAppPath else {
-            return nil
-        }
-        return RealMiniAppView.shared.getMiniAppView(miniAppPath: filePath)
+    func getMiniAppView(miniAppPath: URL) -> MiniAppView {
+        return RealMiniAppView.shared.getMiniAppView(miniAppPath: miniAppPath)
     }
 }
--- a/MiniApp/Classes/Display/Displayer.swift
+++ b/MiniApp/Classes/Display/Displayer.swift
@@ -1,0 +1,9 @@
+internal class Displayer {
+
+    func getMiniAppView(miniAppPath: URL?) -> MiniAppView? {
+        guard let filePath = miniAppPath else {
+            return nil
+        }
+        return RealMiniAppView.shared.getMiniAppView(miniAppPath: filePath)
+    }
+}
--- a/MiniApp/Classes/Display/MiniAppWebView.swift
+++ b/MiniApp/Classes/Display/MiniAppWebView.swift
@@ -4,6 +4,7 @@ internal class MiniAppWebView: WKWebView {
 
     convenience init(filePath: URL) {
         let config = WKWebViewConfiguration()
+        config.preferences.javaScriptEnabled = true
         self.init(frame: .zero, configuration: config)
         contentMode = .scaleToFill
         loadFileURL(filePath, allowingReadAccessTo: filePath.deletingLastPathComponent())
--- a/MiniApp/Classes/Display/MiniAppWebView.swift
+++ b/MiniApp/Classes/Display/MiniAppWebView.swift
@@ -1,2 +1,11 @@
-internal class MiniAppWebView {
+import WebKit
+
+internal class MiniAppWebView: WKWebView {
+
+    convenience init(filePath: URL) {
+        let config = WKWebViewConfiguration()
+        self.init(frame: .zero, configuration: config)
+        contentMode = .scaleToFill
+        loadFileURL(filePath, allowingReadAccessTo: filePath.deletingLastPathComponent())
+    }
 }
--- a/MiniApp/Classes/Display/MiniAppWebView.swift
+++ b/MiniApp/Classes/Display/MiniAppWebView.swift
@@ -2,7 +2,10 @@ import WebKit
 
 internal class MiniAppWebView: WKWebView {
 
-    convenience init(filePath: URL) {
+    convenience init?(filePath: URL) {
+        if !filePath.isFileURL {
+            return nil
+        }
         let config = WKWebViewConfiguration()
         config.preferences.javaScriptEnabled = true
         self.init(frame: .zero, configuration: config)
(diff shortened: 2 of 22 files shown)


MiniApp/Classes/Display/MiniAppWebView.swift
```diff
@@ -1,2 +1,11 @@
-internal class MiniAppWebView {
+import WebKit
+
+internal class MiniAppWebView: WKWebView {
+
+    convenience init(filePath: URL) {
+        let config = WKWebViewConfiguration()
```
corycaywood: Is JavaScript enabled by default, or does it need to be enabled manually? Maybe it's better to explicitly set it to enabled?

For Android WebViews, JS is disabled by default.

MiniApp/Classes/Display/RealMiniAppView.swift
```diff
@@ -1,2 +1,8 @@
-internal class RealMiniAppView: MiniAppView {
+internal class RealMiniAppView {
+    static let shared =  RealMiniAppView()
+
+    func getMiniAppView(miniAppPath: URL) -> MiniAppView {
+        let viewFrame = CGRect(x: 0, y: 0, width: UIScreen.main.bounds.width, height: UIScreen.main.bounds.height)
```
corycaywood: This will set the view to the full screen size? I'm not sure if this is what we want because the user might want to make it a different size. 

Will the user still be able to change the size of the view? Or should we force the user to define the view frame instead of creating one for them? What do you think is best way to give the user flexibility and ease of use? Sorry, I'm not very familiar with how iOS views work...

MiniApp/Classes/Display/RealMiniAppView.swift
```diff
@@ -1,2 +1,8 @@
-internal class RealMiniAppView: MiniAppView {
+internal class RealMiniAppView {
```
corycaywood: I think this class is missing unit tests.

MiniApp/Classes/Display/MiniAppWebView.swift
ts-leojoseph-r: Done

MiniApp/Classes/Display/RealMiniAppView.swift
ts-leojoseph-r: I revisited and made few changes. MiniAppView is of UIView type, so the user can now set the frame using .frame(). I guess we can remove this code.

MiniApp/Classes/Display/RealMiniAppView.swift
ts-leojoseph-r: Done

MiniApp/Classes/Display/Displayer.swift
```diff
@@ -0,0 +1,9 @@
+internal class Displayer {
+
+    func getMiniAppView(miniAppPath: URL?) -> MiniAppView? {
```
ts-nitish-a-jain: I think `miniAppPath` doesn't need to be optional and optionally return MiniAppView. What do you think?

Tests/Unit/MiniAppViewTests.swift
```diff
@@ -0,0 +1,29 @@
...
+                    expect(miniAppView).toEventually(beAnInstanceOf(MiniAppView.self))
+                }
+                it("will return same frame sizes") {
+                    guard let testFileURL = MockFile.createTestFile(fileName: "MiniApp.txt") else {
+                        return
+                    }
+                    let viewFrame = CGRect(x: 0, y: 0, width: UIScreen.main.bounds.width, height: UIScreen.main.bounds.height)
+                    let miniAppView = MiniAppView(filePath: testFileURL)
+                    miniAppView.frame = viewFrame
+                    expect(miniAppView.frame).toEventually(equal(viewFrame))
```
imaginaris: Does it make sense to test that? I'll pass immediately.
Maybe something like `expect(miniAppView.frame.equalTo(viewFrame)).toEventuallyNot(beFalse())`?

MiniApp/Classes/Display/Displayer.swift
ts-leojoseph-r: Initial idea was to have a non-optional URL passed to displayer, but since we didn't have any code written to check the validity of mini path URL, this was added. The other PR is going take care of these check. Removed now. Thanks for the comment

Tests/Unit/MiniAppViewTests.swift
ts-leojoseph-r: Totally. But I wanted to use the API's from the Quick and Nimble as much as possible to compare any objects(at least in the testcases)

imaginaris: Yeah, unfortunately there is no other way in this case. If you want to keep it the way it is now then ok

MiniApp/Classes/Display/MiniAppWebView.swift
```diff
@@ -1,2 +1,12 @@
...
+        config.preferences.javaScriptEnabled = true
+        self.init(frame: .zero, configuration: config)
+        contentMode = .scaleToFill
+        loadFileURL(filePath, allowingReadAccessTo: filePath.deletingLastPathComponent())
```
donnie-jp: We'll use this to isolate the miniapps?

Tests/Unit/DisplayerTests.swift
```diff
@@ -0,0 +1,21 @@
+import Quick
+import Nimble
+@testable import MiniApp
+
+class DisplayerTests: QuickSpec {
+
+    override func spec() {
+        describe("get mini app view") {
+            context("when valid URL of a mini app is passed") {
```
donnie-jp: What happens when the URL is invalid?

Tests/Unit/MiniAppViewTests.swift
```diff
@@ -0,0 +1,29 @@
+import Quick
+import Nimble
+@testable import MiniApp
+
+class MiniAppViewTests: QuickSpec {
+
+    override func spec() {
+        describe("Mini App view") {
+            context("when initialized with valid parameters") {
```
donnie-jp: What about invalid parameters?

Tests/Unit/MiniAppViewTests.swift
donnie-jp: I don't think "will return same frame sizes" tests anything and can be removed.

Tests/Unit/RealMiniAppViewTests.swift
```diff
@@ -0,0 +1,20 @@
+import Quick
+import Nimble
+@testable import MiniApp
+
+class RealMiniAppViewTests: QuickSpec {
+
+    override func spec() {
+        describe("Mini App view") {
+            context("when initialized with valid parameters") {
```
donnie-jp: What about invalid?
Unit tests should cover the "unhappy" paths as well as the "happy" paths

MiniApp/Classes/Display/MiniAppWebView.swift
ts-leojoseph-r: Yes

Tests/Unit/DisplayerTests.swift
ts-leojoseph-r: I need to use failable initializers then. Working on it

Tests/Unit/MiniAppViewTests.swift
ts-leojoseph-r: I need to use failable initializers then. Working on it

Tests/Unit/DisplayerTests.swift
ts-leojoseph-r: Done

Tests/Unit/MiniAppViewTests.swift
ts-leojoseph-r: Done

PR #22, opened 2020-02-17
Updated createMiniApp method with interfaces

ts-leojoseph-r:
# Description
Updated public interface for creating Mini app

# Checklist

### DIFF
--- a/MiniApp/Classes/Display/MiniAppWebView.swift
+++ b/MiniApp/Classes/Display/MiniAppWebView.swift
@@ -3,8 +3,8 @@ import WebKit
 internal class MiniAppWebView: WKWebView {
 
     convenience init?(filePath: URL) {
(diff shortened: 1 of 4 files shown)


MiniApp/Classes/Display/MiniAppWebView.swift
```diff
@@ -3,13 +3,14 @@ import WebKit
 internal class MiniAppWebView: WKWebView {
 
     convenience init?(filePath: URL) {
+        let fileURL = filePath.appendingPathComponent(Constants.FileNames.baseFile)
```
corycaywood: I'm wondering if maybe `index.html` should just be inline here instead of in a constants file? I can't think of any situation right now where anything other than `MiniAppWebView` needs to know about `index.html`.

ts-leojoseph-r: 🤔 Could be. I was thinking like a standard practise to have constants for file names

PR #28, opened 2020-02-20
Updated API endpoints and specs

ts-leojoseph-r:
# Description
Updated API endpoints and spec

# Checklist

### DIFF
--- a/Example/ViewController.swift
+++ b/Example/ViewController.swift
@@ -30,8 +30,7 @@ class ViewController: UITableViewController {
         let cell = tableView.dequeueReusableCell(withIdentifier: "MiniAppCell", for: indexPath)
 
         let miniAppDetail = self.decodeResponse?[indexPath.row]
-        cell.textLabel?.text = miniAppDetail?.name
-        cell.detailTextLabel?.text = miniAppDetail?.description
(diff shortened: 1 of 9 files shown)


Example/ViewController.swift
```diff
@@ -30,8 +30,7 @@ class ViewController: UITableViewController {
         let cell = tableView.dequeueReusableCell(withIdentifier: "MiniAppCell", for: indexPath)
 
         let miniAppDetail = self.decodeResponse?[indexPath.row]
-        cell.textLabel?.text = miniAppDetail?.name
-        cell.detailTextLabel?.text = miniAppDetail?.description
```
ts-nitish-a-jain: Will the detail text label auto-hide when a value is not assigned?

ts-leojoseph-r: Not auto-hide, there will be no value assigned(empty string) as it is optional

ts-nitish-a-jain: Okay. On the Android side, we just modified the visibility to reuse the layout for the future (but we made sure that description text is not visible, to reflect that there have been updates on the BE API). You might want to consider the same.

ts-leojoseph-r: Pushed the changes now.

PR #29, opened 2020-02-25
Develop

Climbatize:
# Description
Added MiniApp informations retrieval the same way we retrieve MiniApp list. As it uses the same API, I kept existing code and added a miniAppId parameter to relevant methods

## Links
[MINI-447] iOS - Load a Mini App by ID

# Checklist
- [ ] I have read the [contributing guidelines](../CONTRIBUTING.md). I read them too late :(
- [ ] I ran `fastlane ci` without errors fastlane CI is running


MiniApp/Classes/MiniAppDownloader.swift
```diff
@@ -27,7 +27,7 @@ class MiniAppDownloader {
             switch result {
             case .success(let responseData):
                 self.completionHandler = completionHandler
-                self.downloadManifestFiles(with: appId, versionId: versionId, files: responseData.manifest)
+                self.downloadManifestFiles(with: appId, versionId: versionId, files: responseData.files)
```
ts-leojoseph-r: As per the latest API specs, it will be `manifest`

MiniApp/Classes/Models/ManifestResponse.swift
```diff
@@ -1,4 +1,4 @@
 /// Mini App Manifest information
 internal struct ManifestResponse: Decodable {
-    var manifest: [String]
+    var files: [String]
```
ts-leojoseph-r: As per the latest API specs, it will be `manifest`

MiniApp/Classes/MiniApp.swift
```diff
@@ -12,6 +12,16 @@ public class MiniApp: NSObject {
         return RealMiniApp.shared.listMiniApp(completionHandler: completionHandler)
     }
 
+    /// Fetch the MiniAppInfo information for a given MiniApp id.
+    ///
+    /// - Parameters:
```
ts-leojoseph-r: Parameters - `miniAppId` detail is missing

MiniApp/Classes/MiniAppLister.swift
```diff
@@ -14,4 +14,20 @@ internal class MiniAppLister {
             }
         }
     }
+
+    func fetchMiniApp(miniAppId: String, apiClient: MiniAppClient, completionHandler: @escaping (Result<MiniAppInfo, Error>) -> Void) {
+
+        apiClient.getMiniApp(miniAppId) { (result) in
+            switch result {
+            case .success(let responseData):
+                guard let decodeResponse = ResponseDecoder.decode(decodeType: Array<MiniAppInfo>.self, data: responseData.data), let miniApp = decodeResponse.first else {
```
ts-leojoseph-r: Array<MiniAppInfo>.self ? I think it will return Mini App info for a given App ID, not an array ?

Tests/Unit/ListingAPITests.swift
```diff
@@ -12,16 +12,26 @@ class ListingAPITests: QuickSpec {
...
             context("when endpoint is properly configured") {
-                it("will return valid URL Request") {
+                it("will return valid URL Request for app listing") {
                     mockBundle.mockEndpoint = "http://example.com"
                     expect(listingAPI.createURLRequest()).toEventually(beAnInstanceOf(URLRequest.self))
                 }
+
+                it("will return valid URL Request for app info") {
+                    mockBundle.mockEndpoint = "http://example.com"
+                    expect(listingAPI.createURLRequest("123")).toEventually(beAnInstanceOf(URLRequest.self))
```
ts-leojoseph-r: Do you think we need to add an extra test case to check whether it will return URLRequest with AppID added in the request?

MiniApp/Classes/MiniAppDownloader.swift
donnie-jp: Can you use codingKeys so that we can call it what we want internally and if the backend keeps changing the naming manifest/files/whatever it won't affect the SDK?

Tests/Unit/MiniAppListerTests.swift
```diff
@@ -103,5 +103,85 @@ class MiniAppListerTests: QuickSpec {
...
+                            decodedResponse = responseData
+                        case .failure:
+                            break
+                        }
+                    })
+                    expect(decodedResponse).toEventually(beAnInstanceOf(MiniAppInfo.self))
+                }
+            }
+            context("when request from server returns invalid data") {
+                it("will decode the response with MiniAppInfo decodable") {
```
donnie-jp: Is this named correctly? Won't the decoding fail?

Tests/Unit/MiniAppListerTests.swift
```diff
@@ -103,5 +103,85 @@ class MiniAppListerTests: QuickSpec {
...
+                        case .success:
+                            break
+                        case .failure(let error):
+                            testError = error as NSError
+                        }
+                    })
+                    expect(testError?.code).toEventually(equal(123))
+                }
+            }
+        }
```
donnie-jp: Maybe these tests can be refactored to reduce the duplication?

MiniApp/Classes/MiniAppLister.swift
Climbatize: no, as per documentation it will return an array of one element

donnie-jp:
Why was this PR closed and a new PR#30 created for the same changes?

PR #30, opened 2020-02-25
Feature/447

Climbatize:
## Description

Added MiniApp informations retrieval the same way we retrieve MiniApp list. As it uses the same API, I kept existing code and added a miniAppId parameter to relevant methods

## Links
[MINI-447] iOS - Load a Mini App by ID

# Checklist


MiniApp/Classes/API/ListingAPI.swift
```diff
@@ -5,17 +5,26 @@ internal class ListingApi {
...
     }
 
-    private func getListingURL() -> URL? {
+    private func getListingURL(_ miniAppId: String? = nil) -> URL? {
         guard let baseURL = environment.baseUrl else {
             return nil
         }
-        return baseURL.appendingPathComponent("/oneapp/ios/\(environment.appVersion)/miniapps")
+
+        guard var components = URLComponents(url: baseURL.appendingPathComponent("/oneapp/ios/\(environment.appVersion)/miniapps"), resolvingAgainstBaseURL: false) else {
```
ts-nitish-a-jain: I think `/oneapp/` is not there in the latest spec. Can you please have a look?

MiniApp/Classes/API/ListingAPI.swift
```diff
@@ -5,17 +5,26 @@ internal class ListingApi {
         self.environment = environment
     }
 
-    func createURLRequest() -> URLRequest? {
-        guard let url = getListingURL() else {
+    func createURLRequest(miniAppID: String? = nil) -> URLRequest? {
```
ts-nitish-a-jain: `Optional`?

MiniApp/Classes/MiniAppLister.swift
```diff
@@ -14,4 +14,21 @@ internal class MiniAppLister {
...
+    func fetchMiniApp(miniAppId: String, apiClient: MiniAppClient, completionHandler: @escaping (Result<MiniAppInfo, Error>) -> Void) {
+
+        apiClient.getMiniApp(miniAppId) { (result) in
+            switch result {
+            case .success(let responseData):
+                guard let decodeResponse = ResponseDecoder.decode(decodeType: Array<MiniAppInfo>.self, data: responseData.data), let miniApp = decodeResponse.first else {
+                    return completionHandler(.failure(NSError.invalidResponseData()))
+                }
+
+                return completionHandler(.success(miniApp))
```
ts-nitish-a-jain: Two different styles of spacing/formatting around `return`, one here and one above.

MiniApp/Classes/API/ListingAPI.swift
ts-nitish-a-jain: My bad. This is for another PR.

MiniApp/Classes/API/ListingAPI.swift
```diff
@@ -5,17 +5,26 @@ internal class ListingApi {
...
+        guard let url = getListingURL(miniAppID) else {
             return nil
         }
         return URLRequest.createURLRequest(url: url, environment: environment)
     }
 
-    private func getListingURL() -> URL? {
+    private func getListingURL(_ miniAppId: String? = nil) -> URL? {
```
ts-nitish-a-jain: Can this be broken into smaller units, I see multiple `returns`?

MiniApp/Classes/MiniApp.swift
```diff
@@ -12,6 +12,17 @@ public class MiniApp: NSObject {
...
 
+    /// Fetch the MiniAppInfo information for a given MiniApp id.
+    ///
+    /// - Parameters:
+    ///     -   miniAppId: the identifier string of the Mini App you want information
+    ///     -   completionHandler: A block to be called when MiniAppInfo information is fetched. Completion blocks receives the following parameters
+    ///         -   MiniAppInfo: MiniAppInfo information.
+    ///         -   Error: Error details if fetching is failed.
+    public class func info(miniAppId: String, completionHandler: @escaping (Result<MiniAppInfo, Error>) -> Void) {
+        return RealMiniApp.shared.getMiniApp(miniAppId: miniAppId, completionHandler: completionHandler)
```
ts-nitish-a-jain: I think it will be good to place validations at the public API for empty inputs of appId. What do you think?

MiniApp/Classes/MiniAppLister.swift
```diff
@@ -14,4 +14,21 @@ internal class MiniAppLister {
             }
         }
     }
+
+    func fetchMiniApp(miniAppId: String, apiClient: MiniAppClient, completionHandler: @escaping (Result<MiniAppInfo, Error>) -> Void) {
```
ts-nitish-a-jain: `fetchMiniApp` sounds like retrieving the complete Mini App whereas the function is about getting the metadata. 
On Android side, we are have evolved the `List` module into `AppInfo`.

ts-nitish-a-jain: You can refer: https://github.com/rakutentech/android-miniapp/pull/45/files

MiniApp/Classes/API/ListingAPI.swift
Climbatize: yes, as it is in the API call. If I have the parameter I add it to the call to get the listing with only 1 mini app, else I make the regular call for the whole listing

MiniApp/Classes/API/ListingAPI.swift
Climbatize: I did this to keep compatibility with existing code. The new parameter miniAppId is optional and nil if not provided

MiniApp/Classes/MiniApp.swift
Climbatize: you might be right, I'll fix this

MiniApp/Classes/MiniAppLister.swift
Climbatize: OK, sounds good to me

MiniApp/Classes/API/ListingAPI.swift
```diff
@@ -1,21 +1,47 @@
...
             return nil
         }
-        return baseURL.appendingPathComponent("/oneapp/ios/\(environment.appVersion)/miniapps")
+        return baseURL.appendingPathComponent(path)
+    }
+
+    private func getListingURL(for miniAppId: String) -> URL? {
+        guard let baseURL = environment.baseUrl else {
+            return nil
+        }
```
ts-leojoseph-r: Maybe we can move this into some function to avoid duplicates?

Climbatize: yes, got it

PR #31, opened 2020-02-25
Saving download status of mini app using userdefaults

ts-leojoseph-r:
# Description
Download status of Mini app is saved now using UserDefaults (created with suiteName) and skipping the download if mini-app is already downloaded

# Checklist


Tests/Unit/MiniAppPreferencesTests.swift
```diff
@@ -0,0 +1,29 @@
+import Quick
+import Nimble
+@testable import MiniApp
+
+class MiniAppPreferencesTests: QuickSpec {
+
+    override func spec() {
```
Climbatize: I think it's not necessary in this case, but should we mock Userdefaults contained in MiniAppPreferences for unit testing by using UserDefaults().removePersistentDomain(forName: "com.rakuten.tech.mobile.miniapp")

MiniApp/Classes/Storage/MiniAppPreferences.swift
```diff
@@ -0,0 +1,15 @@
+class MiniAppPreferences {
```
corycaywood: How about the name `MiniAppStatus` instead? This matches the name on Android and more clearly communicates what this class does.

ts-leojoseph-r: Done

PR #33, opened 2020-03-02
Refactoring download mini app operation

ts-leojoseph-r:
# Description
Moved existing download operation logic into a new class for two reasons below,
* Avoid completion handler memory leak
* To return the Mini App path once the downloading is complete.

# Checklist

### DIFF
--- a/MiniApp/Classes/DownloadOperation.swift
+++ b/MiniApp/Classes/DownloadOperation.swift
@@ -1,70 +1,13 @@
 class DownloadOperation {
     typealias DownloadCompletionHandler = (Result<URL, Error>) -> Void
 
-    private var completionHandler: DownloadCompletionHandler?
-    private var urlToDirectoryMap = [String: URL]()
-    private var miniAppClient: MiniAppClient
-    private var miniAppStorage: MiniAppStorage
-    private var miniAppPath: URL
+    var fileStoragePath: URL
+    var miniAppDirectoryPath: URL
+    var completionHanlder: DownloadCompletionHandler
 
-    private var queue: OperationQueue = {
-        let operationQueue = OperationQueue()
-        operationQueue.name = "MiniAppDownloader"
-        operationQueue.maxConcurrentOperationCount = 4
-        return operationQueue
-    }()
-
-    init(miniAppStorage: MiniAppStorage, miniAppClient: MiniAppClient, miniAppPath: URL, completionHandler: @escaping DownloadCompletionHandler) {
-        self.miniAppClient = miniAppClient
-        self.miniAppStorage = miniAppStorage
-        self.miniAppPath = miniAppPath
-        self.completionHandler = completionHandler
-    }
-
-    func downloadFiles(urls: [String]) {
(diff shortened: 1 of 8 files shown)


MiniApp/Classes/DownloadOperation.swift
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    func cancelAllOperations() {
+        queue.cancelAllOperations()
+    }
+}
+
+extension DownloadOperation: MiniAppDownloaderProtocol {
+
+    func fileDownloaded(sourcePath: URL, destinationPath: String) {
```
ts-nitish-a-jain: Can we rename the methods to be descriptive?
https://swift.org/documentation/api-design-guidelines/#promote-clear-usage

MiniApp/Classes/DownloadOperation.swift
```diff
@@ -0,0 +1,70 @@
...
+
+    func downloadFiles(urls: [String]) {
+        self.miniAppClient.delegate = self
+        for url in urls {
+            guard let fileDirectory = UrlParser.parseForFileDirectory(with: url) else {
+                self.completionHandler?(.failure(NSError.downloadingFailed()))
+                return
+            }
+            if !FileManager.default.fileExists(atPath: self.miniAppPath.appendingPathComponent(fileDirectory).path) {
+                urlToDirectoryMap[url] = self.miniAppPath.appendingPathComponent(fileDirectory)
```
ts-nitish-a-jain: Should this happen after a successful download of that file instead?

ts-leojoseph-r: We are also using this value capture to track the download progress that's why we are saving before the download starts

MiniApp/Classes/DownloadOperation.swift
ts-leojoseph-r: Updating now.

MiniApp/Classes/DownloadOperation.swift
```diff
@@ -0,0 +1,70 @@
+class DownloadOperation {
```
corycaywood: Looks like there aren't any unit tests for this class?

MiniApp/Classes/DownloadOperation.swift
ts-nitish-a-jain: Is it getting reset if the particular download fails?

MiniApp/Classes/DownloadOperation.swift
ts-leojoseph-r: Any suggestions for the method name 🤔

MiniApp/Classes/DownloadOperation.swift
ts-leojoseph-r: DownloadOperation object itself will be released once the download completes/fails

MiniApp/Classes/DownloadOperation.swift
ts-leojoseph-r: Working on it

PR #34, opened 2020-03-03
UI to test MiniApp info fetch by ID

Climbatize:
# Description
Adds a way to call Mini App Info API via test user interface

## Links
MINI-447

# Checklist


Example/ViewController.swift
```diff
@@ -22,6 +23,39 @@ class ViewController: UITableViewController {
         }
     }
 
+    @IBAction func actionShowMiniApp() {
```
corycaywood: Maybe `actionShowMiniAppById` instead so as to not get confused with showing a Mini App when clicking on the list item?

Climbatize: you're right @corycaywood , I pushed a refactor

Example/ViewController.swift
```diff
@@ -22,6 +23,39 @@ class ViewController: UITableViewController {
         }
     }
 
+    @IBAction func actionShowMiniAppById() {
+        let alert = UIAlertController(title: "Please enter Mini App ID",
```
ts-nitish-a-jain: Looks like a long method, on the Android side we get complaints about something over 15 lines.
And, possibly extracting the messages into a single place. It helps to keep things tidy. What do you think?

Climbatize: mmm well, it's the shortest way to create such a dialog. I could extract several methods from this, or make an extension if you think it's valuable for a demo  
As for the messages, we could use localizable strings to do things properly, what do you think about this? By the way this makes code a little bit more verbose

Example/ViewController.swift
```diff
@@ -22,6 +23,39 @@ class ViewController: UITableViewController {
...
+                        self.performSegue(withIdentifier: "DisplayMiniApp", sender: nil)
+                    case .failure(let error):
+                        print(error.localizedDescription)
+                        self.displayErrorAlert(title: "Error", message: "Couldn't retrieve Mini App info, please try again later", dismissController: false)
+                    }
+                    self.dismissProgressIndicator()
+                }
+            } else {
+                self.dismissProgressIndicator()
+                self.displayErrorAlert(title: "Error", message: "Incorrect Mini App ID, please try again", dismissController: false)
```
ts-leojoseph-r: As of now, this might not display an error alert message "Incorrect ..." because dimissProgressIndicator will dismiss displayErrorAlert immediately. I faced a weird issue.
You can remove `self.dismissProgressIndicator()` and then you can set the `dismissController: true` in `self.displayErrorAlert()`

Example/ViewController.swift
```diff
@@ -22,6 +23,39 @@ class ViewController: UITableViewController {
...
+                                      preferredStyle: .alert)
+
+        alert.addTextField { (textField) in
+            textField.keyboardType = .asciiCapable
+        }
+
+        let okAction = UIAlertAction(title: "OK", style: .default) { (_) in
+            self.showProgressIndicator()
+            if let textField = alert.textFields?.first, let miniAppID = textField.text, miniAppID.count > 0 {
+                MiniApp.info(miniAppId: miniAppID) { (result) in
```
ts-leojoseph-r: Maybe we can move this block to a separate method?

Example/ViewController.swift
ts-nitish-a-jain: Overall, it helps to keep the code tidy. Of course, for the demo app sometimes we too make some trade-offs.

I think a minor refactor (usually less than <20 mins) is good in general, the next person might feel glad to have found some pieces in place when doing similar stuff. What do you think?
btw, localizable strings sound good to me.

MiniApp/en.lproj/Localizable.strings
```diff
@@ -0,0 +1,19 @@
+/* 
+  Localizable.strings
+  MiniApp
+
+  Created by Villa, PierreMichel on 2020/03/04.
+  Copyright © 2020 CocoaPods. All rights reserved.
+*/
```
ts-leojoseph-r: Please remove the boiler plate

Example/ViewController.swift
```diff
@@ -17,7 +18,38 @@ class ViewController: UITableViewController {
...
+                                    self.performSegue(withIdentifier: "DisplayMiniApp", sender: nil)
+                                case .failure(let error):
+                                    print(error.localizedDescription)
+                                    self.displayErrorAlert(
+                                        title: NSLocalizedString("error_title", comment: ""),
+                                        message: NSLocalizedString("error_single_message", comment: ""),
+                                        dismissController: false)
+                                }
+                            }
+                        }
```
ts-leojoseph-r: Maybe we can move this into a separate function?

MiniApp/en.lproj/Localizable.strings
Climbatize: done

Example/ViewController.swift
Climbatize: also done, but I don't find it simplifies reading that much :)

ts-leojoseph-r: It was a suggestion, I felt that the method looked little bigger and possibility that it can be broken down

ts-nitish-a-jain: some useful stuff: https://arturbosch.github.io/detekt/complexity.html#longmethod

MiniApp/en.lproj/Localizable.strings
```diff
@@ -0,0 +1,11 @@
+"error_title" = "Error";
+"error_list_message" = "Couldn't retrieve Mini App list, please try again later";
+"error_single_message" = "Couldn't retrieve Mini App info, please try again later";
+"error_incorrect_appid_message" = "Incorrect Mini App ID, please try again";
+"error_miniapp_message" = "Couldn't retrieve Mini App, please try again later";
+
+"input_miniapp_title" = "Please enter Mini App ID";
+
+"wait_message" = "Please wait...";
+"OK" = "OK";
```
ts-nitish-a-jain: I think the `key` can still be lower case 😄

Climbatize: OK 😂

PR #35, opened 2020-03-05
Change App ID and Subscription Key at runtime

Climbatize:
# Description
**Add feature for changing the App ID and Subscription Key at runtime**: For this I added a MiniAppSdkConfig the same way as we did on Android. It is optional and can be omitted if we want the bundle values . 
**Also add a settings dialog in the Demo App for changing the App ID and Subscription Key. The Demo App should cache the App ID/Key when it is changed so that the cached values are still used when the app is re-launched** : I added a tab bar in demo app in which there is a config screen where the user can override all the values of MiniAppSdkConfig

## Links
MINI-625

# Checklist


Pods/Local Podspecs/MiniApp.podspec.json
```diff
@@ -0,0 +1,19 @@
+{
```
ts-leojoseph-r: Do you think its a good idea to commit Pod files?
Please check https://guides.cocoapods.org/using/using-cocoapods.html#should-i-ignore-the-pods-directory-in-source-control

MiniApp/Classes/MiniApp.swift
```diff
@@ -5,36 +5,83 @@ public class MiniApp: NSObject {
...
     ///                         to interact with View component of mini app.
     ///         -   Error: Error details if Mini App View creating is failed
-    public class func create(appInfo: MiniAppInfo, completionHandler: @escaping (Result<MiniAppDisplayProtocol, Error>) -> Void) {
-        return RealMiniApp.shared.createMiniApp(appInfo: appInfo, completionHandler: completionHandler)
+    public class func create(config: MiniAppSdkConfig? = nil, appInfo: MiniAppInfo, completionHandler: @escaping (Result<MiniAppDisplayProtocol, Error>) -> Void) {
+        return RealMiniApp.shared(with: config).createMiniApp(appInfo: appInfo, completionHandler: completionHandler)
+    }
+}
+
+public class MiniAppSdkConfig {
```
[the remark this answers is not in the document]
Climbatize: Yes we can. I just followed  what was done on Android. By the way, this class is key to the new feature, this is the config you provide to MiniApp calls to override manifest parameters. So yes, it has to be public.

MiniApp/Classes/MiniAppClient.swift
```diff
@@ -84,14 +92,14 @@ class MiniAppClient: NSObject, URLSessionDownloadDelegate {
...
                                                           data: responseData) else {
-                                                            return NSError.unknownServerError(httpResponse: httpResponse)
+                return NSError.unknownServerError(httpResponse: httpResponse)
             }
             message = "\(errorModel.error): \(errorModel.errorDescription)"
         default:
             guard let errorModel = ResponseDecoder.decode(decodeType: ErrorData.self,
                                                           data: responseData) else {
-                    return NSError.unknownServerError(httpResponse: httpResponse)
-                }
```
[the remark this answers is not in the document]
Climbatize: Unwanted spaces on new lines were making me crazy because Xcode automatically adds spaces on new lines and SwiftLint don't want them, so I installed a plungin in Xcode that reformat the code so it is 100% swift lint compliant. These were bad formatted lines so I thought it would make no harm if I committed them

Pods/Local Podspecs/MiniApp.podspec.json
Climbatize: Well, Cocoapods recommendation from the page you linked is :

> Whether or not you check in your Pods folder is up to you, as workflows vary from project to project. We recommend that you keep the Pods directory under source control, and don't add it to your .gitignore. But ultimately this decision is up to you

Personally I'm not advising to commit pods, because it's a lot of files that can be retrieved easily with pod install. The fact is this documentation page is not taking in account the case were you are working on a new pod. What happens if you don't commit your development pod files is that if one of the team members adds a new file in the pod structure, this file won't be added to the pod project structure of your teammates when they merge your changes... unless you pod update, what I totally missed ! So the answer is no, it's not a good idea, but it's not a bad idea either. I think I'll remove them from source control anyway and stick to the pod update way.

MiniApp/Classes/MiniApp.swift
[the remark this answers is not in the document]
ts-nitish-a-jain: On Android, we are trying to figure out other probable issues. I think let's keep in close sync w.r.t. to those. 
I also agree that this should be in a separate class. And since it is public, it should have some documentation too. 
@corycaywood Let's move it to a separate class as well in Android.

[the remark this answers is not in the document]
Climbatize: I moved it

MiniApp/Classes/MiniApp.swift
```diff
@@ -5,36 +5,39 @@ public class MiniApp: NSObject {
     /// Error information will be returned if any problem while fetching from the backed
     ///
     /// - Parameters:
+    ///     -   config: on optional custom configuration to override default SDK parameters such as base URL, app ID, subsciption key or host app version
```
ts-nitish-a-jain: In the public API documentation, I think it will be good to specify the use cases when to use this optional config e.g. QA (it gives SDK authors a good chance to specify the intended use as well).

MiniApp/Classes/MiniApp.swift
```diff
@@ -5,36 +5,39 @@ public class MiniApp: NSObject {
...
     /// - Parameters:
+    ///   - config: on optional custom configuration to override default SDK parameters such as base URL, app ID, subsciption key or host app version
     ///   - appInfo: Mini App info object
     ///   - completionBlock: A block to be called on successful creation of [MiniAppView] or throws errors if any. Completion blocks receives the following parameters
     ///         -   MiniAppDisplayProtocol: Protocol that helps the hosting application to communicate with the displayer module of the mini app. More like an interface for host app
     ///                         to interact with View component of mini app.
     ///         -   Error: Error details if Mini App View creating is failed
-    public class func create(appInfo: MiniAppInfo, completionHandler: @escaping (Result<MiniAppDisplayProtocol, Error>) -> Void) {
-        return RealMiniApp.shared.createMiniApp(appInfo: appInfo, completionHandler: completionHandler)
+    public class func create(config: MiniAppSdkConfig? = nil, appInfo: MiniAppInfo, completionHandler: @escaping (Result<MiniAppDisplayProtocol, Error>) -> Void) {
```
ts-nitish-a-jain: I was looking at the Android PR before this. In iOS, I see that using `MiniApp.swift` the functions are being accessed statically on the app side. On the Android side, even though a single instance is being maintained for this, it is still accessed via the object. @minh-rakuten leveraged this and has handled `config` differently. It looks like a nice idea, what do you think @Climbatize ?
It has been working fine as the public APIs were more or less the same for both the platforms. Even after this, a small deviation shouldn't impact as such 💭

Pods/Local Podspecs/MiniApp.podspec.json
ts-leojoseph-r: You are right. We can skip this for now, but we should find a workaround for this.

MiniApp/Classes/MiniApp.swift
ts-nitish-a-jain: @ts-leojoseph-r it is only a single instance with ability to update the configuration.

Climbatize: I moved the singleton from RealMiniApp to MiniApp to match Android

MiniApp/Classes/MiniAppSdkConfig.swift
```diff
@@ -0,0 +1,45 @@
...
+        }
+    }
+    var hostAppVersion: String? {
+        didSet(value) {
+            if value?.count ?? 0 == 0 {
+                hostAppVersion = nil
+            }
+        }
+    }
+    var key: String {
```
corycaywood: On android, the `key` is being used to store different instances of the api client. But on iOS, there's only the single instance of the api client, so I think this `key` isn't being used for anything, right?

MiniApp.xcodeproj/project.pbxproj
```diff
@@ -7,7 +7,12 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		0B9872320F362A7239F5FC0A /* Pods_MiniApp_Tests.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = E6D517CF1E4DE7AA981E7496 /* Pods_MiniApp_Tests.framework */; };
 		181DC083240F316D00FDF633 /* Localizable.strings in Resources */ = {isa = PBXBuildFile; fileRef = 181DC085240F316D00FDF633 /* Localizable.strings */; };
+		18C59F2B241094F600F66632 /* SettingsTableViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 18C59F2A241094F600F66632 /* SettingsTableViewController.swift */; };
+		18C59F2D2410ADE300F66632 /* Config.swift in Sources */ = {isa = PBXBuildFile; fileRef = 18C59F2C2410ADE300F66632 /* Config.swift */; };
+		18D764392419D08600150A30 /* Config.swift in Sources */ = {isa = PBXBuildFile; fileRef = 18C59F2C2410ADE300F66632 /* Config.swift */; };
+		32EBAB652329E85618B76E53 /* Pods_MiniApp_Example.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 9D87B5609C299582369DA550 /* Pods_MiniApp_Example.framework */; };
```
ts-leojoseph-r: Pods files details need to be committed in the git, you can get rid of them by doing `pod deintegrate `

Example/Config.swift
```diff
@@ -0,0 +1,21 @@
...
+
+class Config: NSObject {
+    enum Key: String {
+        case applicationIdentifier = "RASApplicationIdentifier",
+        version = "CFBundleShortVersionString",
+        subscriptionKey = "RASProjectSubscriptionKey",
+        endpoint = "RMAAPIEndpoint"
+    }
+
+    static let userDefaults = UserDefaults(suiteName: "com.rakuten.tech.mobile.miniapp.MiniAppDemo.settings")
```
donnie-jp: Could you initialize the class with a `UserDefaults` object, with default param set to what you've defined here?
I think it would make testing easier.

donnie-jp: Ah, it seems like you need this to be static.

Example/ViewController.swift
```diff
@@ -5,35 +5,48 @@ class ViewController: UITableViewController {
...
+        showProgressIndicator {
+            MiniApp.shared(with: self.config).list { (result) in
+                DispatchQueue.main.async {
+                    self.refreshControl?.endRefreshing()
+                }
+                switch result {
+                case .success(let responseData):
+                    self.decodeResponse = responseData
+                    self.tableView.reloadData()
+                    self.dismissProgressIndicator()
```
donnie-jp: Does it need dismissed in the failure case?

Example/Config.swift
Climbatize: good idea, I'll implement this

MiniApp/Progress+UIViewController.swift
```diff
@@ -20,11 +20,11 @@ extension UIViewController {
         })
     }
 
-    func displayErrorAlert(title: String, message: String, dismissController: Bool) {
+    func displayAlert(title: String, message: String, dismissController: Bool) {
         DispatchQueue.main.asyncAfter(deadline: .now()) {
             self.dismiss(animated: true, completion: {
                 let alert = UIAlertController(title: title, message: message, preferredStyle: .alert)
-                alert.addAction(UIAlertAction(title: NSLocalizedString("OK", comment: ""), style: .cancel, handler: { (_: UIAlertAction!) in
+                alert.addAction(UIAlertAction(title: NSLocalizedString("ok", comment: ""), style: .cancel, handler: { (_: UIAlertAction!) in
```
donnie-jp: Maybe an extension that just lets you do `"ok".localized` would be better.

Climbatize: well, I could, but I think you can't use automatized string extraction anymore (`xcrun extractLocStrings *.swift`)

Example/ViewController.swift
Climbatize: no, because we have the self.displayAlert that manages this. I think these methods have to be refactored by the way

PR #37, opened 2020-03-11
Delete files for old versions

Climbatize:
# Description
After a new version of a Mini App has finished downloading, delete the files for the previously downloaded versions of the Mini App.

## Links
MINI-614

# Checklist


MiniApp/Classes/Extensions/MiniApp+FileManager.swift
```diff
@@ -11,10 +11,21 @@ extension FileManager {
...
+    }
+
+    /*
+     * Provide the MiniApp directory by appending the system
+     * cache directory with AppID.
+     * @param { appId: String } - AppID of the MiniApp.
+     * @param { versionId: String } - VersionID of the MiniApp.
+     * @return { URL } - URL path to the MiniApp.
+     */
+    class func getMiniAppDirectory(with appId: String) -> URL {
```
ts-leojoseph-r: Maybe a different method name? I see two same methods returning different path

MiniApp/Classes/MiniAppDownloader.swift
```diff
@@ -56,6 +65,21 @@ class MiniAppDownloader {
             completionHandler(.success(miniAppPath))
         }
     }
+
+    private func cleanVersions(for appId: String, differentFrom versionId: String) {
```
ts-leojoseph-r: Maybe we can move this to the MiniAppStorage class? What do you think?

MiniApp/Classes/MiniAppDownloader.swift
```diff
@@ -56,6 +65,21 @@ class MiniAppDownloader {
...
+        let miniAppStoragePath = FileManager.getMiniAppDirectory(with: appId)
+
+        if let directoryContents = try? FileManager.default.contentsOfDirectory(at: miniAppStoragePath, includingPropertiesForKeys: nil, options: .skipsSubdirectoryDescendants) {
+            for path in directoryContents where path.lastPathComponent != versionId {
+                do {
+                    try FileManager.default.removeItem(at: path)
+                    miniAppStatus.setDownloadStatus(false, appId: appId, versionId: versionId)
```
ts-leojoseph-r: Guess, we need to update the MiniAppStatus class to remove the value when deleting a specific version. I will update the MiniAppStatus class.

MiniApp/Classes/Extensions/MiniApp+FileManager.swift
```diff
@@ -11,10 +11,21 @@ extension FileManager {
...
      */
     class func getMiniAppDirectory(with appId: String, and versionId: String) -> URL {
+        return getMiniAppDirectory(with: appId).appendingPathComponent("\(versionId)/")
+    }
+
+    /*
+     * Provide the MiniApp directory by appending the system
+     * cache directory with AppID.
+     * @param { appId: String } - AppID of the MiniApp.
+     * @param { versionId: String } - VersionID of the MiniApp.
```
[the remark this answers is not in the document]
Climbatize: oops

MiniApp/Classes/Extensions/MiniApp+FileManager.swift
Climbatize: I rename the other one to getMiniAppVersionDirectory which is more accurate

MiniApp/Classes/MiniAppDownloader.swift
Climbatize: I temporarily managed a way to manage it in MiniAppStorage, see my next PR

MiniApp/Classes/MiniAppDownloader.swift
```diff
@@ -22,15 +22,26 @@ class MiniAppDownloader {
...
         }
         self.manifestDownloader.fetchManifest(apiClient: self.miniAppClient, appId: appId, versionId: versionId) { (result) in
             switch result {
             case .success(let responseData):
-                self.downloadMiniApp(urls: responseData.manifest, to: miniAppStoragePath, completionHandler: completionHandler)
+                self.downloadMiniApp(urls: responseData.manifest, to: miniAppStoragePath) { downloadResult in
+                    switch downloadResult {
+                    case .success:
+                        if let error = self.miniAppStorage.cleanVersions(for: appId, differentFrom: versionId, status: self.miniAppStatus) {
+                            print("WARNING: MiniAppDownloader could not delete previously downloaded versions for appId \(appId) (\(error))")
```
corycaywood: Should you use the `Logger` class for this instead?

MiniApp/Classes/MiniAppDownloader.swift
```diff
@@ -22,15 +22,26 @@ class MiniAppDownloader {
...
             return
         }
         self.manifestDownloader.fetchManifest(apiClient: self.miniAppClient, appId: appId, versionId: versionId) { (result) in
             switch result {
             case .success(let responseData):
-                self.downloadMiniApp(urls: responseData.manifest, to: miniAppStoragePath, completionHandler: completionHandler)
+                self.downloadMiniApp(urls: responseData.manifest, to: miniAppStoragePath) { downloadResult in
+                    switch downloadResult {
+                    case .success:
+                        if let error = self.miniAppStorage.cleanVersions(for: appId, differentFrom: versionId, status: self.miniAppStatus) {
```
ts-nitish-a-jain: Wouldn't this add delay to provide the display unit back to consumer app? 
It would be a good idea to decouple it.

MiniApp/Classes/MiniAppDownloader.swift
Climbatize: we talked about removing it with @leorjoseph so I did not want to use it, but well, seems a good idea if we are keeping it

MiniApp/Classes/MiniAppDownloader.swift
Climbatize: yes if they are working on a potato 🥔 . But in this case, wouldn't it be harder to them to display a miniApp while older versions are deleted (system lagging) ? It's not sarcasm, it's a real question

ts-nitish-a-jain: On a potato? I missed some context. :)

I am not 100% sure how iPhones respond to parallel execution but I do think they have the cores to do it. 😉  A mini app is allowed to be 50MB in size, can easily have thousands of files (like in panda park, disk I/O is expensive).

ts-leojoseph-r: I think it would be a great idea to move them into [Operation](https://developer.apple.com/documentation/foundation/operation) or Dispatch block and immediately return the display unit. Because few mini-apps files/structure would be complex (it could even cause one-second delay) to delete it.

Climbatize: it's not really a threading issue, it's more about memory throttling on old devices

Climbatize: it's not really a threading issue, it's more about memory throttling on old devices

Climbatize: By the way, it doesn't hurt to put it into an async thread ;)

ts-nitish-a-jain: The display doesn't need to be blocked by deletion.

Climbatize: > On a potato? I missed some context. :)
> 
> I am not 100% sure how iPhones respond to **parallel** execution but I do think they have the _cores_ to do it. 😉 A mini app is allowed to be 50MB in size, can easily have thousands of files (like in panda park, disk I/O is expensive).

A potato is a very old device on end of life, just a joke :) Well on very old devices, with cheap/end of life memory, accessing storage memory can be really problematic, and it can be problematic to launch new version of PandaPark while its old version is being deleted. I'm not talking about iPhone X, and you're both right it's better to thread this, but user experience on old device can be degraded far more than waiting a few more seconds.

MiniApp.xcodeproj/project.pbxproj
```diff
@@ -31,8 +31,10 @@
 		38F3B7F024192E7300914E37 /* UrlParserTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 38F3B7EF24192E7300914E37 /* UrlParserTests.swift */; };
 		38F73B4523D554FE00F2C1CE /* Helpers.swift in Sources */ = {isa = PBXBuildFile; fileRef = 38F73B4423D554FE00F2C1CE /* Helpers.swift */; };
 		38F73B4723D55CCB00F2C1CE /* MiniAppInfoFetcherTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 38F73B4623D55CCB00F2C1CE /* MiniAppInfoFetcherTests.swift */; };
+		58A0DF3B3769B8951169516B /* Pods_MiniApp_Tests.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 52C30D2FD93FB4105B5C47EA /* Pods_MiniApp_Tests.framework */; };
```
ts-leojoseph-r: Please avoid pushing this change into GIT, you can do this by pod deintegrate. If these changes are pushed, then for every PR, we have to review this file.

PR #38, opened 2020-03-11
Refactor - Removed logger class and updated few testcases

ts-leojoseph-r:
# Description
* Renamed logger class to MiniAppLogger
* Added test cases for UrlParser class
  * UrlParser will be updated soon with the lastest endpoints changes

# Checklist


Tests/Unit/UrlParserTests.swift
```diff
@@ -0,0 +1,24 @@
...
+                    let urlString = "https://www.example.com/version/rak0123/img/onload/file.png"
+                    let path = UrlParser.parseForFileDirectory(with: urlString)
+                    expect(path).toEventually(equal("img/onload/file.png"))
+                }
+            }
+            context("when parseForFileDirectory is called with invalid url") {
+                it("will return nil") {
+                    let urlString = "https://www.example.com/mini-ver/rak0123/img/onload/file.png"
+                    let path = UrlParser.parseForFileDirectory(with: urlString)
+                    expect(path).toEventuallyNot(equal("img/onload/file.png"))
```
Climbatize: why not `expect(path).toEventually(beNil())` ?

ts-leojoseph-r: Well, the Urlparser doesn't seem to be returning nil for now. Like I said in the PR description, this class need to refactored to update the latest endpoints and most of these things can be avoided.

Climbatize: Oh, I just was confused by `it("will return nil")`, sorry

MiniApp/Classes/Environment.swift
```diff
@@ -44,7 +44,6 @@ internal class Environment {
 
     var baseUrl: URL? {
         guard let endpointUrlString = (self.customUrl ?? bundle.value(for: Key.endpoint.rawValue)) else {
-            Logger.e("Ensure RMAAPIEndpoint value in plist is valid")
```
corycaywood: Should we really delete this? It seems it would be useful for developers. Because the error that's returned when they try to create a Mini App won't be very helpful because it won't mention anything about RMAAPIEndpoint, right?

MiniApp/Classes/Logger.swift
```diff
@@ -1,37 +0,0 @@
...
-extension OSLog {
-    static let sdk = OSLog(category: "Mini App SDK")
-
-    private convenience init(category: String, bundle: Bundle = Bundle(for: Logger.self)) {
-        let identifier = bundle.infoDictionary?["CFBundleIdentifier"] as? String
-        self.init(subsystem: (identifier ?? "").appending(".logs"), category: category)
-    }
-}
-
-internal class Logger {
```
corycaywood: It seems like this class is very helpful for printing debug messages, such as in the above case. Even if it's not being used much right now, maybe it will still be useful?

donnie-jp: Yeah, I don't understand why a configurable utility class like Logger has been deleted.

ts-nitish-a-jain: I also don't see any good reason for removing this from the SDK code.

ts-nitish-a-jain: In fact, should also get branded to be `MiniAppLogger` or alike.

ts-leojoseph-r: Noted. I think we should start using this class more in the SDK. Renaming the class with a new name

PR #41, opened 2020-03-17
Refactor/update endpoints change

ts-leojoseph-r:
# Description
As per the latest API specs, updated the code to use **map-published** string as a separator and retrieve the storage path for the given URL.
Updated test cases for the same

# Checklist

### DIFF
--- a/MiniApp/Classes/Utilities/UrlParser.swift
+++ b/MiniApp/Classes/Utilities/UrlParser.swift
@@ -12,7 +12,7 @@ struct UrlParser {
      * @return { String } - The file directory after removing the base URL.
      */
     static func getFileStoragePath(from url: String) -> String? {
-        if (url.contains(separatorKey)) {
+        if url.contains(separatorKey) {
             var pathComponents: [String] = url.components(separatedBy: separatorKey)
             guard let path: String = pathComponents.last, !path.isEmpty else {
                 MiniAppLogger.e("MiniAppSDK: Failed parsing URL.")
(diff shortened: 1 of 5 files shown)


MiniApp/Classes/Utilities/UrlParser.swift
```diff
@@ -3,20 +3,28 @@
...
      */
-    static func parseForFileDirectory(with url: String) -> String? {
-        guard let versionId =
-            url.components(separatedBy: "version/").last?.components(separatedBy: "/").first,
-            let directory = url.components(separatedBy: "\(versionId)/").last
-        else {
-            MiniAppLogger.e("MiniAppSDK: Failed parsing URL.")
-            return nil
+    static func getFileStoragePath(from url: String) -> String? {
+        if url.contains(separatorKey) {
```
ts-nitish-a-jain: `contains` & `components` might lead to the same logical outcome (making the lookup by `contains` redundant). What do you think?

Climbatize: I was thinking the same as you @ts-nitish-a-jain at first, but if the string does not contain the separator, the pathComponents will still have one element. So it's not the same logic. Maybe pathComponents size could be tested.

ts-leojoseph-r: Yes, @Climbatize is right. Even if components couldn't find an element (provided with separator) it will still return an element.

PR #44, opened 2020-03-26
Feature/javascript interface base

ts-leojoseph-r:
# Description
This is the base code for adding the MiniAppMessageProtocol method into the SDK. 

Note: Actual implementation will be done in a separate PR

# Checklist

### DIFF
--- a/Example/DisplayController.swift
+++ b/Example/DisplayController.swift
@@ -1,7 +1,8 @@
 import UIKit
 import MiniApp
 
-class DisplayController: UIViewController {
+class DisplayController: UIViewController, MiniAppMessageProtocol {
+    
     var miniAppInfo: MiniAppInfo?
     var config: MiniAppSdkConfig?
 
@@ -15,7 +16,7 @@ class DisplayController: UIViewController {
             return
         }
         self.title = info.displayName
-        MiniApp.shared(with: config).create(appInfo: info) { (result) in
+        MiniApp.shared(with: config).create(appInfo: info, completionHandler: { (result) in
             switch result {
             case .success(let miniAppDisplay):
                 self.dismissProgressIndicator()
@@ -26,6 +27,10 @@ class DisplayController: UIViewController {
                 self.displayAlert(title: NSLocalizedString("error_title", comment: ""), message: NSLocalizedString("error_miniapp_download_message", comment: ""), dismissController: true)
                 print("Errored: ", error.localizedDescription)
             }
-        }
+        }, messageInterface: self)
+    }
+
+    func getUniqueId() -> String {
+        return ""
     }
 }
--- a/Example/DisplayController.swift
+++ b/Example/DisplayController.swift
@@ -2,7 +2,7 @@ import UIKit
 import MiniApp
 
 class DisplayController: UIViewController, MiniAppMessageProtocol {
-    
+
     var miniAppInfo: MiniAppInfo?
     var config: MiniAppSdkConfig?
 
--- a/Example/DisplayController.swift
+++ b/Example/DisplayController.swift
@@ -7,14 +7,15 @@ class DisplayController: UIViewController, MiniAppMessageProtocol {
     var config: MiniAppSdkConfig?
 
     override func viewWillAppear(_ animated: Bool) {
-        self.showProgressIndicator()
         config = Config.getCurrent()
     }
 
     override func viewDidAppear(_ animated: Bool) {
-        guard let info = miniAppInfo else {
+        guard let controller = self.navigationController as? DisplayNavigationController, let info = controller.miniAppInfo else {
             return
         }
+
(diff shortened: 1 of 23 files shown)


Example/DisplayController.swift
```diff
@@ -26,6 +27,10 @@ class DisplayController: UIViewController {
                 self.displayAlert(title: NSLocalizedString("error_title", comment: ""), message: NSLocalizedString("error_miniapp_download_message", comment: ""), dismissController: true)
                 print("Errored: ", error.localizedDescription)
             }
-        }
+        }, messageInterface: self)
+    }
+
+    func getUniqueId() -> String {
+        return ""
```
Climbatize: shouldn't we return some dummy value ? Or make it nullable ?

MiniApp/Classes/RealMiniApp.swift
```diff
@@ -31,7 +31,7 @@ internal class RealMiniApp {
         return miniAppInfoFetcher.getInfo(miniAppId: miniAppId, apiClient: self.miniAppClient, completionHandler: completionHandler)
     }
 
-    func createMiniApp(appInfo: MiniAppInfo, completionHandler: @escaping (Result<MiniAppDisplayProtocol, Error>) -> Void) {
+    func createMiniApp(appInfo: MiniAppInfo, completionHandler: @escaping (Result<MiniAppDisplayProtocol, Error>) -> Void, messageInterface: MiniAppMessageProtocol) {
```
corycaywood: I think that the `messageInterface` needs to be passed to `displayer.getMiniAppView` so that it can be used from the `WKScriptMessageHandler`, right?

Example/DisplayController.swift
corycaywood: Yeah, we should make it return some value. Does iOS have a unique ID (either unique to the device or App) you can use here?

MiniApp/Classes/RealMiniApp.swift
ts-leojoseph-r: Done

Example/DisplayController.swift
ts-leojoseph-r: Done

MiniApp/Classes/Display/Displayer.swift
```diff
@@ -1,6 +1,6 @@
 internal class Displayer {
 
-    func getMiniAppView(miniAppId: String) -> MiniAppDisplayProtocol? {
-        return RealMiniAppView.shared.getMiniAppView(miniAppId: miniAppId)
+    func getMiniAppView(miniAppId: String, messageInterface: MiniAppMessageProtocol) -> MiniAppDisplayProtocol? {
```
ts-nitish-a-jain, line 3: Is the message interface essential for getting the view from the display?

ts-leojoseph-r: Yes, because javascript interface methods will be implemented in `MiniAppView` class, so we have to pass this object.

ts-nitish-a-jain: In Android, @minh-rakuten has created a separate method to allow injection via the contract: `injectJSInterface`. I think we will need to align a bit as it is Public API.

MiniApp/Classes/Protocols/MiniAppMessageProtocol.swift
```diff
@@ -0,0 +1,7 @@
+/**
+Public Protocol that will be used by the Mini App to communicate
+ with the Native implementation
+*/
+public protocol MiniAppMessageProtocol {
+    func getUniqueId() -> String
```
ts-nitish-a-jain, line 6: `getUniqueId()`seems to have missed the doc.

MiniApp/Classes/Display/Displayer.swift
ts-leojoseph-r: I don't understand why we need to have a contract here?  This isn't something that will be done on the native side right?

ts-nitish-a-jain: I think this one looks correct. Emphasizes the need/dependency on the consumer.

MiniApp/Classes/Protocols/MiniAppMessageProtocol.swift
```diff
@@ -0,0 +1,9 @@
+/**
+Public Protocol that will be used by the Mini App to communicate
+ with the Native implementation
+*/
+public protocol MiniAppMessageProtocol {
+
+    /// Interface that should be implemented to return alphanumeric string that uniquely identifies a device.
```
ts-nitish-a-jain, line 7: Was it the device or installation 🤔 ?

ts-leojoseph-r: Both 😉 One device one installation

PR #45, opened 2020-04-01
Updated testcases for RealMini app

ts-leojoseph-r:
# Description
Created test cases for Real Mini App class to increase the code coverage

# Checklist

### DIFF
--- a/Tests/Unit/Helpers.swift
+++ b/Tests/Unit/Helpers.swift
@@ -160,5 +160,8 @@ var mockMiniAppInfo: MiniAppInfo {
 ///   - appId: Mini App ID
 func deleteMockMiniApp(appId: String, versionId: String) {
     try? FileManager.default.removeItem(at: FileManager.getMiniAppDirectory(with: appId))
-    UserDefaults(suiteName: "com.rakuten.tech.mobile.miniapp")?.set(false, forKey: "\(appId)/\(versionId)")
+}
+
+func deleteStatusPreferences() {
+    UserDefaults.standard.removePersistentDomain(forName: "com.rakuten.tech.mobile.miniapp")
 }
--- a/Tests/Unit/Helpers.swift
+++ b/Tests/Unit/Helpers.swift
@@ -148,3 +148,17 @@ class MockFile {
         return fullURL
     }
 }
+
+var mockMiniAppInfo: MiniAppInfo {
+    let mockVersion = Version(versionTag: "Dev", versionId: "ver-id-test")
+    let info = MiniAppInfo.init(id: "app-id-test", displayName: "Mini App Title", icon: URL(string: "https://www.example.com/icon.png")!, version: mockVersion)
+    return info
+}
+
+/// Method to delete the Mini App directory which was created for Mock testing
+/// - Parameters:
+///   - appId: Mini App ID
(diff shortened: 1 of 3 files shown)


Tests/Unit/Helpers.swift
```diff
@@ -148,3 +148,17 @@ class MockFile {
...
+    let info = MiniAppInfo.init(id: "app-id-test", displayName: "Mini App Title", icon: URL(string: "https://www.example.com/icon.png")!, version: mockVersion)
+    return info
+}
+
+/// Method to delete the Mini App directory which was created for Mock testing
+/// - Parameters:
+///   - appId: Mini App ID
+func deleteMockMiniApp(appId: String, versionId: String) {
+    try? FileManager.default.removeItem(at: FileManager.getMiniAppDirectory(with: appId))
+    UserDefaults(suiteName: "com.rakuten.tech.mobile.miniapp")?.set(false, forKey: "\(appId)/\(versionId)")
```
Climbatize, line 163: We could also remove the preference for the key

Tests/Unit/RealMiniAppTests.swift
```diff
@@ -0,0 +1,116 @@
...
+                    let realMiniApp = RealMiniApp()
+                    let mockAPIClient = MockAPIClient()
+                    let mockManifestDownloader = MockManifestDownloader()
+                    let downloader = MiniAppDownloader(apiClient: mockAPIClient, manifestDownloader: mockManifestDownloader, status: miniAppStatus)
+                    realMiniApp.miniAppClient = mockAPIClient
+                    realMiniApp.manifestDownloader = mockManifestDownloader
+                    realMiniApp.miniAppDownloader = downloader
+                    let responseString = """
+                      {
+                        "manifest": ["https://google.com/map-published/app-id-test/ver-id-test/HelloWorld.txt"]
```
Climbatize, line 71: should we have a Rakuten domain name instead :)

ts-leojoseph-r: Changed to generic name example.com.

Tests/Unit/Helpers.swift
ts-leojoseph-r: Done

Tests/Unit/RealMiniAppTests.swift
```diff
@@ -0,0 +1,117 @@
...
+        describe("Real mini app tests") {
+            afterEach {
+                deleteMockMiniApp(appId: mockMiniAppInfo.id, versionId: mockMiniAppInfo.version.versionId)
+                deleteStatusPreferences()
+            }
+            context("when getMiniApp is called with valid app id") {
+                it("will return valid MiniAppInfo") {
+                    let realMiniApp = RealMiniApp()
+                    let mockAPIClient = MockAPIClient()
+                    realMiniApp.miniAppClient = mockAPIClient
```
corycaywood, line 19: It seems you need to create and set the mocks on each test. Could you do this kind of setup in `beforeEach` instead of in each test?

ts-leojoseph-r: That makes sense, but beforeEach is a closure and the objects declared in it can't be used outside. So I had moved those repeated codes to global now.

PR #46, opened 2020-04-06
Removed MiniAppView and moved the code to RealMiniAppView

ts-leojoseph-r:
# Description
Removed MiniAppView class and moved the logic to RealMiniAppView to sync up with Architecture

# Checklist


Tests/Unit/RealMiniAppViewTests.swift
```diff
@@ -2,14 +2,21 @@ import Quick
 import Nimble
 @testable import MiniApp
 
-class RealMiniAppViewTests: QuickSpec {
```
ts-nitish-a-jain, line 5: Should the swift file name also change?

ts-leojoseph-r: Done

PR #56, opened 2020-04-15
Feature/mini 858

Climbatize:
# Description
fix : Can open unpublished version as long as the list isn't refreshed. To avoid this I load MiniApp before launching WKWebview controller.

## Links
MINI-858

# Checklist


Climbatize:
> Overall Looks good to me 👍
> I see fetching lists, create mini-app info and fetch by app info all moved to one place which is good. I feel that maybe we should think of moving extensions to separate file to make it more clear. Maybe later.
> 
> Please resolve the conflicts so I approve the PR

done

PR #57, opened 2020-04-16
Optional value set for displayName in MiniAppInfo

ts-leojoseph-r:
# Description
* Changes are done to display the App name as "Null" if API returns displayName as null.
* Removed SF_list_dash imageset from xcassets (image was removed but imageset isn't)

## Links
[MINI-825
](https://jira.rakuten-it.com/jira/browse/MINI-825)
# Checklist


Example/ViewController.swift
```diff
@@ -88,6 +88,9 @@ class ViewController: UITableViewController {
         if let cell = tableView.dequeueReusableCell(withIdentifier: "MiniAppCell", for: indexPath) as? MiniAppCell {
             let miniAppDetail = self.decodeResponse?[indexPath.row]
             cell.titleLabel?.text = miniAppDetail?.displayName
+            if (miniAppDetail?.displayName) == nil {
+                cell.titleLabel?.text = "Null"
```
Climbatize, line 92: what do you think of
`cell.titleLabel?.text = miniAppDetail?.displayName ?? "Null"`

ts-leojoseph-r: Done

PR #58, opened 2020-04-16
ci: Add pod publishing via tag on travis ci

corycaywood:
# Description
Pod will be published when a tagged is pushed that begins with 'v'

# Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,6 @@ xcode_scheme: Tests
 osx_image: xcode11.2
 
 before_install:
-- bundle install
-- gem update fastlane cocoapods --no-document
 - pod repo update
 
 script:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -20,14 +18,25 @@ after_success:
...
 
+deploy:
+  provider: script
+  skip_cleanup: true
+  script: |
+    pod spec lint --allow-warnings 
+    && pod trunk push --allow-warnings --verbose
+  on:
+    tags: true
+    condition: $TRAVIS_TAG =~ ^v
```
donnie-jp, line 29: https://guides.cocoapods.org/making/making-a-cocoapod.html#library-versioning suggests that prefixing x.y.z numbers with `v` isn't standard and I'm not sure it's supported by CocoaPods dependency resolver. Can you confirm?

corycaywood: I've also updated the `source` to prepend `v` to the tag. I found this solution here: https://stackoverflow.com/a/22646192
I also tested this by adding a tag on my fork and updating the podspec to point to my fork. After doing this, `pod spec lint` passed. Does this mean it should work?

corycaywood: Also, I saw that Cocoapod's examples use `v` in the tag: https://guides.cocoapods.org/syntax/podspec.html

PR #59, opened 2020-04-16
Display error message when a non uuid (mini app) id format is entered

ts-leojoseph-r:
# Description
* Validate if the entered mini-app id is valid format(UUID format with hyphens) and display inline error message in AlertController

## Links
[MINI-576](https://jira.rakuten-it.com/jira/browse/MINI-576)
# Checklist


Example/Progress+UIViewController.swift
```diff
@@ -71,4 +80,19 @@ extension UIViewController {
...
+    ///   - message: Error message description
+    func displayErrorInAlertController(alertController: UIAlertController, message: String) {
+        let paragraphStyle = NSMutableParagraphStyle()
+        paragraphStyle.alignment = NSTextAlignment.center
+        let attributedString = NSAttributedString(string: message, attributes: [
+            NSAttributedString.Key.paragraphStyle: paragraphStyle,
+            NSAttributedString.Key.font: UIFont.systemFont(ofSize: 14),
+            NSAttributedString.Key.foregroundColor: UIColor.red
+        ])
+        alertController.setValue(attributedString, forKey: "attributedMessage")
```
donnie-jp, line 96: Using Apple private API?
Hmm. I'm not sure - maybe it's ok for the sample app that will never be released.

ts-leojoseph-r: Yeah, I was thinking to create a Custom Alert button but thought of using private API to make the job easier for the sample app.

Climbatize: I couldn't agree more with @donnie-jp, especially if it's just to have a custom colored message blinking in the alert :)
But I have to admit it is nicely done and Apple will never review this app, so... Let's just hope Apple don't refactor its AlertController in a near future. ;)

ts-leojoseph-r: Totally agreed. Let me if I can add some other alternative approach

ts-leojoseph-r: I guess the other alternatives also end up in modifying UIAlertController(Which we shouldn't). So I guess we can go ahead with this solution

Climbatize: Yes, the main purpose of this PR is to filter bad UUID, which is done properly. We can merge this PR and change the controller if something goes wrong in our future tests. A simple change would be to have a regular UIViewController instead of an UIAlertViewController

PR #60, opened 2020-04-17
Readme file

Climbatize:
# Description
Updated readme file

## Links
MINI-869


README.md
```diff
@@ -16,17 +34,55 @@ Generated documentation for this SDK can be found at https://rakutentech.github.
 
 To run the example project, clone the repo, and run `pod install` from the Example directory first.
 
-## Requirements
+### Custom parameters
+
+In order to load your MiniApp, you will have to use your own Host App ID and your subscription key. These can eiher be set in project configuration plist (RASApplicationIdentifier and RASProjectSubscriptionKey) or by taping the top right configuration icon in the example application.
```
corycaywood, line 39: Need to also mention that they should set the base url. On android, we also have a message like this: `We don't currently host a public API, so you will need to provide your own Base URL for API requests.`.

README.md
```diff
@@ -16,17 +34,55 @@ Generated documentation for this SDK can be found at https://rakutentech.github.
...
+MiniApp is available through [CocoaPods](https://cocoapods.org). To install it, simply add the following line to your Podfile:
 
 ```ruby
 pod 'MiniApp'
 ```
 
+In your project configuration .plist you can add 2 custom string target properties :
+
+- `RASApplicationIdentifier` - to set your MiniApp host application identifier
+- `RASProjectSubscriptionKey` - to set your MiniApp subsbscription key
```
corycaywood, line 52: Also should list the base url property here.

README.md
```diff
@@ -6,7 +6,25 @@
 
 # MiniApp
 
-This SDK provides capabilities to downloading/cache and host the Mini-App in native iOS applications. It also provides more features to interact with the native application via bridge.
+This open-source library allows you to integrate a Rakuten MiniApp into your iOS applications. It also provides more features to interact with the native application via bridge.
+
+## Features
+
+- Load MiniApp list
+- Load MiniApp infos
```
ts-leojoseph-r, line 14: Maybe information or meta-information ?

README.md
```diff
@@ -6,7 +6,25 @@
...
+
+## Getting started
+
+### Requirements
+
+This module supports iOS 11.0 and above. It has been tested on iOS 11.0 and above.
+
+It is written in Swift 5.0 and can be used in compatible Xcode versions.
+
+In order to run your MiniApp you will have to provide your MiniApp host application identifier and subsbscription key to the library
```
ts-leojoseph-r, line 27: Subscription key?

README.md
```diff
@@ -16,17 +34,55 @@ Generated documentation for this SDK can be found at https://rakutentech.github.
...
+MiniApp.shared().info(miniAppId: miniAppID) { (result) in
+	...
+}
+```
+
+Create a MiniApp for the given `MiniAppInfo` object :
+
+```
+MiniApp.shared().create(appInfo: info, completionHandler: { (result) in
+	...
```
corycaywood, line 82: I think the example for `create` should be a little more complete.  On Android side, there is mostly complete Activity example showing how they would set the mini app to the view after retrieving it.

README.md
ts-leojoseph-r: I guess Endpoint detail is missing here

README.md
ts-leojoseph-r: Agree with cory + eiher -> either

README.md
```diff
@@ -16,17 +34,55 @@ Generated documentation for this SDK can be found at https://rakutentech.github.
...
+
+If you don't want to use project settings, you are still able to pass this informations one by one to the `Config.userDefaults` using a `Config.Key` as key:
+
+```
+Config.userDefaults?.set("MY_CUSTOM_ID", forKey: Config.Key.subscriptionKey.rawValue)
+```
+
+## Usage
+
+MiniApp library calls are done via the `MiniApp.shared()` singleton with or without a `MiniAppSdkConfig` instance (you can get the current one with `Config.getCurrent()`). If you don't provide a config instance, values in custiom iOS target properties will be used by default. For example, if you want to load the MiniApp list:
```
ts-leojoseph-r, line 62: custiom -> custom?

README.md
```diff
@@ -16,17 +34,55 @@ Generated documentation for this SDK can be found at https://rakutentech.github.
...
+
+```
+MiniApp.shared(with: Config.getCurrent()).list { (result) in
```
ts-leojoseph-r, line 65: Do you think its a good idea to add another approach as well i.e w/o passing the Config

README.md
Climbatize: 😅

README.md
Climbatize: You're right, I base the doc on the custom runtime parameters we had in the application

README.md
Climbatize: yes it is 👍

README.md
Climbatize: done

ts-leojoseph-r:
> Oh, sorry I just noticed something else. Can you also check the public interface docs?
> 
> I noticed that `RealMiniAppView` is public but it should be private. Most of the public interface has documentation, but please check that all of it does. I think `MiniAppInfo` is missing documentation.

Thanks for spotting it @corycaywood. Done now

PR #62, opened 2020-04-22
Validations for AppID and Subscription key - MINI-578

ts-leojoseph-r:
# Description
* Changes to validate Mini App ID and Subscription key in the Settings UI
* Enable/Disable Save button only when the user enters a value in TextField
* Fetch mini-app list only when valid host app id and subscription entered in Custom Config
* Save custom configuration only when valid Host app id and subscription entered
* Enable/Disable 'OK' Button in 'Load Mini app by ID' flow

## Links
[MINI-578](https://jira.rakuten-it.com/jira/browse/MINI-578)

# Checklist

### DIFF
--- a/Example/Base.lproj/Main.storyboard
+++ b/Example/Base.lproj/Main.storyboard
@@ -84,7 +84,7 @@
                         <barButtonItem key="backBarButtonItem" title=" " id="6gZ-4N-NPm"/>
                         <barButtonItem key="rightBarButtonItem" title="Item" image="Settings" id="QKl-NE-KE1">
                             <connections>
-                                <segue destination="TVZ-Rp-2Bg" kind="presentation" modalPresentationStyle="fullScreen" id="6nU-Qi-B5w"/>
+                                <segue destination="TVZ-Rp-2Bg" kind="presentation" identifier="CustomConfiguration" id="6nU-Qi-B5w"/>
                             </connections>
                         </barButtonItem>
                     </navigationItem>
@@ -110,7 +110,7 @@
                 </tableViewController>
                 <placeholder placeholderIdentifier="IBFirstResponder" id="Zsb-zf-A01" userLabel="First Responder" customClass="UIResponder" sceneMemberID="firstResponder"/>
             </objects>
-            <point key="canvasLocation" x="229.59999999999999" y="385.45727136431788"/>
+            <point key="canvasLocation" x="-50" y="358"/>
         </scene>
         <!--Mini Apps-->
         <scene sceneID="oLG-1t-f3c">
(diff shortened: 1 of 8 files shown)


Example/String+MiniApp.swift
```diff
@@ -0,0 +1,12 @@
+import Foundation
+
+extension String {
+
+    func isValidUUID() -> Bool {
+        if(self.range(of: #"[0-9a-f]{8}-?[0-9a-f]{4}-?[0-9a-f]{4}-?[0-9a-f]{4}-?[0-9a-f]{12}+$"#,
```
corycaywood, line 6: You're using Regex because `UUID()` must be passed a string containing hypens? As far as I'm aware, the `UUID` class on Android behaves the same way.

Do we actually need to support the case when the UUID doesn't have hyphens? I think that the UUID's we're using will always have hyphens. It just seems like this Regex adds a lot of complexity for a case that we probably don't need to handle.

ts-leojoseph-r: @corycaywood 
We had a similar approach like Android but that was validating only host app ID.

I added this logic for two purposes, 
1. HostAppId - UUID Type
    - UUID String (with hyphens)
2. Subscription key
    - String with the same length as UUID but just without hyphens

I thought it would be good to have a unique function that tells whether it is valid HostAppID (UUID) and Subscription key.
This function is not exactly meant to validate UUID (w or w/o hyphens) but to validate HostAppId and subscription key

corycaywood: Ah, I understand now. But the subscription key isn't a UUID, so I don't think it should be validated in the same way. Actually, I don't think we should do any validation on the subscription key because we can't predict what format it will be. 

Also, this change means that validation for the host app id won't be correct, right? Because the host app id is always supposed to have hyphens.

ts-leojoseph-r: It can validate both.

Ok, I will revert back the change related to the old style that will validate Host app ID (UUID type) and Subscription key (just make sure it is not empty) 👍

Example/Base.lproj/Main.storyboard
```diff
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="16096" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" colorMatched="YES" initialViewController="hzK-gd-cH8">
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="15702" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" colorMatched="YES" initialViewController="hzK-gd-cH8">
```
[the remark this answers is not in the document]
ts-leojoseph-r: 👍

PR #63, opened 2020-04-23
Updating to latest JS bridging code

ts-leojoseph-r:
# Description
* Updating Bridging script
    * Bridging script is updated after test cases and formatting changes

# Checklist

### DIFF
--- a/MiniApp/Classes/JavascriptBridge/bridge.js
+++ b/MiniApp/Classes/JavascriptBridge/bridge.js
@@ -1,13 +1,14 @@
-var messageQueue = []
-window.MiniAppBridge = {};
+var MiniAppBridge = {};
+MiniAppBridge.messageQueue = [];
 var uniqueId = Math.random();
+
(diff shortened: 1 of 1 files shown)


MiniApp/Classes/JavascriptBridge/bridge.js
```diff
@@ -1,13 +1,14 @@
...
 var uniqueId = Math.random();
+
 var isPlatform = {
-    Android: function() {
-        return navigator.userAgent.match(/Android/i);
-    },
-    iOS: function() {
-        return navigator.userAgent.match(/iPhone|iPad|iPod/i);
-    }
+  Android: function() {
```
Climbatize, line 6: do we ever call isPlatform.Android() somewhere ?

ts-leojoseph-r: Not from iOS but the same generic bridging script has been used by Android too right?

PR #68, opened 2020-04-27
Feature/MINI-834

Climbatize:
# Description
While we want to show developers how to initialize the SDK at build time, the injection of these parameters confuses the experience when opening the app the first time by potentially showing Apps that aren't the users. 
To solve this issue I added an initial controller which controls if parameters have been set.

## Links
MINI-834

# Checklist


Example/SettingsTableViewController.swift
```diff
@@ -162,3 +162,7 @@ class SettingsTableViewController: UITableViewController {
         self.navigationItem.rightBarButtonItem?.isEnabled = true
     }
 }
+
+protocol SettingsTableViewControllerDelegate: class {
+    func settingsTableViewController(_ controller: SettingsTableViewController, updated miniAppList: [MiniAppInfo])
```
ts-leojoseph-r, line 167: Maybe a different protocol method name? I feel Protocol name and the method name are much alike

Example/FirstLaunchViewController.swift
```diff
@@ -0,0 +1,47 @@
...
+        self.performSegue(withIdentifier: "showList", sender: nil)
+    }
+
+    override func prepare(for segue: UIStoryboardSegue, sender: Any?) {
+        if segue.identifier == "CustomConfiguration" {
+            if let navigationController = segue.destination as? UINavigationController,
+                let customSettingsController = navigationController.topViewController as? SettingsTableViewController {
+                customSettingsController.configUpdateDelegate = self
+            }
+        } else if segue.identifier == "showList" {
```
ts-leojoseph-r, line 35: Camel case for segue identifiers?

Example/FirstLaunchViewController.swift
```diff
@@ -0,0 +1,47 @@
...
+        guard Config.userDefaults?.string(forKey: Config.Key.applicationIdentifier.rawValue) != nil, Config.userDefaults?.string(forKey: Config.Key.subscriptionKey.rawValue) != nil else {
+            UIView.animate(withDuration: 1.5) {
+                self.imageViewArrow.alpha = 1.0
+                self.labelHint.alpha = 1.0
+            }
+            UIView.animate(withDuration: 0.5) {
+                self.labelMiniApp.alpha = 0.0
+            }
+            return
+        }
```
ts-leojoseph-r, line 25: Maybe we can move this to a separate function?

Example/SettingsTableViewController.swift
Climbatize: that's because the delegate should always have methods recalling its name, like tableView delegate methods for example. Maybe I could simplify it. What do you think ?

ts-leojoseph-r: Yeah, I agree. But I see the protocol name seems to be big and maybe we can have some other prefix like SettingsUpdate/ConfigUpdate or anything.

Example/FirstLaunchViewController.swift
Climbatize: your prefer Pascal ? I'll update

Example/FirstLaunchViewController.swift
Climbatize: sure

Example/SettingsTableViewController.swift
Climbatize: yep, I'll think about it, thanks for noticing this

Climbatize: done

Example/FirstLaunchViewController.swift
Climbatize: done

Example/FirstLaunchViewController.swift
Climbatize: done

PR #73, opened 2020-05-12
[BUGFIX] Can't load Mini App by ID in demo 

Climbatize:
# Description
Developer can't open his MiniApp by ID in demo. Nasty bug, easy fix.

## Links
MINI-888

# Checklist
- [] I wrote/updated tests for new/changed code

### DIFF
--- a/Example/ViewController+MiniApp.swift
+++ b/Example/ViewController+MiniApp.swift
@@ -45,7 +45,6 @@ extension ViewController {
                         self.fetchMiniAppUsingId(title: NSLocalizedString("error_title", comment: ""), message: NSLocalizedString("error_single_message", comment: ""))
                     }
                 }
-                self.dismissProgressIndicator()
             }
         }
     }


Example/ViewController+MiniApp.swift
```diff
@@ -45,7 +45,6 @@ extension ViewController {
                         self.fetchMiniAppUsingId(title: NSLocalizedString("error_title", comment: ""), message: NSLocalizedString("error_single_message", comment: ""))
                     }
                 }
-                self.dismissProgressIndicator()
```
ts-leojoseph-r, line 48: I don't understand why this is blocking load by mini-app functionality? I can still load mini-app by ID

Climbatize: MiniApp is loaded, controller is presented and dismissed even before it appeared, because e `dismissProgressIndicator` only does a UIViewController `self.dismiss` method call which dismisses the presented controller.

PR #80, opened 2020-05-29
Added Fixed Display Mini app by ID button in header view

ts-leojoseph-r:
# Description
Display Mini app by ID button should be a static cell at the top and it needs to be displayed even when the user scrolls the list.
Removed the cell from the Storyboard tableview and added it as HeaderView to remain static at the top.

## Links
[MINI-963](https://jira.rakuten-it.com/jira/browse/MINI-963)

# Checklist


ts-leojoseph-r:
> A little bit overkill but, well, it’s working 😁

I don't see any other possible solution, other than changing the ViewController to conform to UITableView Protocol

PR #92, opened 2020-06-15
Revert back changes related to First launch flow

ts-leojoseph-r:
# Description
* Reverted back changes on how to show the first launch screen
* Changes to sync up with Android behavior
    * Table view will not fetch the list until the config is saved for the first time.

# Checklist

### DIFF
--- a/Example/SettingsTableViewController.swift
+++ b/Example/SettingsTableViewController.swift
@@ -8,6 +8,7 @@ class SettingsTableViewController: UITableViewController {
     @IBOutlet weak var invalidHostAppIdLabel: UILabel!
     @IBOutlet weak var invalidSubscriptionKeyLabel: UILabel!
     weak var configUpdateDelegate: SettingsDelegate?
+    let predefinedKeys: [String] = ["RAS_APPLICATION_IDENTIFIER", "RAS_SUBSCRIPTION_KEY", ""]
 
     override func viewDidLoad() {
         super.viewDidLoad()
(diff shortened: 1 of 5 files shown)


Example/SettingsTableViewController.swift
```diff
@@ -107,11 +108,19 @@ class SettingsTableViewController: UITableViewController {
...
-        field?.text = Config.userDefaults?.string(forKey: key.rawValue)
+        field?.text = getTextFieldValue(key: key, placeholderText: field?.placeholder)
+    }
+
+    func getTextFieldValue(key: Config.Key, placeholderText: String?) -> String? {
+        if predefinedKeys.contains(placeholderText ?? "") {
+            return Config.userDefaults?.string(forKey: key.rawValue) ?? ""
+        } else {
+            return placeholderText
+        }
```
Climbatize, line 119: is `predefinedKeys` still useful ? What can be the case when `predefinedKeys.contains(placeholderText ?? "")` is `false` ?

ts-leojoseph-r: In the first launch scenario, we need to get the values from the plist and assign those text to the text field's. 
I can do this in [code](https://github.com/rakutentech/ios-miniapp/blob/master/Example/SettingsTableViewController.swift#L110) like below,
`field?.text = Config.userDefaults?.string(forKey: key.rawValue) ?? field?.placeholder` 

but if we do so, then the values of the textfields will be 'RAS_APPLICATION_IDENTIFIER' and 'RAS_SUBSCRIPTION_KEY'. Hence to avoid this scenario I added a small workaround

Climbatize: in fact `field?.text = Config.userDefaults?.string(forKey: key.rawValue)` may be enough, but nvm, you did a good job, it's working well ;)

PR #93, opened 2020-06-16
Feature/offline loading

ts-leojoseph-r:
# Description
Added a missing flow for Offline load support. Offline mode will work as follows,
* Get the mini-app info from the host app and check if it is the latest version available. If there is a latest version available, download it else download the version which is passed from the host-app
* If the device is offline and it will lookup for the mini-app info is downloaded already and if it is the same will be returned.
* If the mini-info is not downloaded already, it will return the previously cached version of the mini-app.

## Links
[MINI-1019](https://jira.rakuten-it.com/jira/browse/MINI-1019)

# Checklist

### DIFF
--- a/MiniApp/Classes/MiniAppDownloader.swift
+++ b/MiniApp/Classes/MiniAppDownloader.swift
@@ -38,23 +38,12 @@ class MiniAppDownloader {
                          }
                      }
                  case .failure(let error):
-                    self.handleDownloadError(appId: appId, error: error, completionHandler: completionHandler)
+                    completionHandler(.failure(error))
                  }
              }
         } else {
             completionHandler(.success(miniAppStoragePath))
         }
-    }
-
-    func handleDownloadError(appId: String, error: Error, completionHandler: @escaping (Result<URL, Error>) -> Void) {
-        let downloadError = error as NSError
-        if Constants.offlineErrorCodeList.contains(downloadError.code) {
-           guard let miniAppPath = self.getCachedMiniApp(appId: appId) else {
-               return completionHandler(.failure(downloadError))
-           }
-           completionHandler(.success(miniAppPath))
-        }
-        return completionHandler(.failure(error))
     }
 
     func isMiniAppAlreadyDownloaded(appId: String, versionId: String) -> Bool {
@@ -70,17 +59,27 @@ class MiniAppDownloader {
 
     /// Check if there is any old version of mini app cached for a given app ID
     /// - Parameter appId: App ID for which the version directory path to be fetched
+    /// - Parameter versionId: Version ID of the app info
     /// - Returns: URL of the available cached version for a given mini app ID
-    func getCachedMiniApp(appId: String) -> URL? {
-        guard let versionDirectory = FileManager.getMiniAppVersionDirectory(usingAppId: appId) else {
+    func getCachedMiniAppVersion(appId: String, versionId: String) -> String? {
+        if !isMiniAppAlreadyDownloaded(appId: appId, versionId: versionId) {
+            let cachedVersion = self.miniAppStatus.getCachedVersion(key: appId)
+            if !cachedVersion.isEmpty {
+                var isDirectory: ObjCBool = true
+                let miniAppPath = FileManager.getMiniAppVersionDirectory(with: appId, and: cachedVersion)
+                if FileManager.default.fileExists(atPath: miniAppPath.path, isDirectory: &isDirectory) {
+                    return cachedVersion
+                }
+            }
             return nil
-        }
-        var isDirectory: ObjCBool = true
-        if FileManager.default.fileExists(atPath: versionDirectory.path, isDirectory: &isDirectory) {
-            return versionDirectory
         } else {
-            return nil
+            let versionDirectory = FileManager.getMiniAppVersionDirectory(with: appId, and: versionId)
+            var isDirectory: ObjCBool = true
+            if FileManager.default.fileExists(atPath: versionDirectory.path, isDirectory: &isDirectory) {
+                return versionId
+            }
         }
+        return nil
     }
 
     private func startDownloadingFiles(urls: [String], to miniAppPath: URL, completionHandler: @escaping (Result<URL, Error>) -> Void) {
--- a/MiniApp/Classes/MiniAppDownloader.swift
+++ b/MiniApp/Classes/MiniAppDownloader.swift
@@ -23,26 +23,55 @@ class MiniAppDownloader {
 
     func download(appId: String, versionId: String, completionHandler: @escaping (Result<URL, Error>) -> Void) {
         let miniAppStoragePath = FileManager.getMiniAppVersionDirectory(with: appId, and: versionId)
-        if miniAppStatus.isDownloaded(appId: appId, versionId: versionId) {
+        if !isMiniAppDownloadedAlready(appId: appId, versionId: versionId) {
+            self.manifestDownloader.fetchManifest(apiClient: self.miniAppClient, appId: appId, versionId: versionId) { (result) in
+                 switch result {
+                 case .success(let responseData):
+                     self.downloadMiniApp(urls: responseData.manifest, to: miniAppStoragePath) { downloadResult in
+                         switch downloadResult {
+                         case .success:
+                             self.miniAppStorage.cleanVersions(for: appId, differentFrom: versionId, status: self.miniAppStatus)
+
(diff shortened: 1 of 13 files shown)


MiniApp/Classes/MiniAppDownloader.swift
```diff
@@ -23,27 +23,56 @@ class MiniAppDownloader {
...
+    }
 
-                        fallthrough
-                    default:
-                        completionHandler(downloadResult)
-                    }
-                }
-            case .failure(let error):
-                return completionHandler(.failure(error))
+    func isMiniAppDownloadedAlready(appId: String, versionId: String) -> Bool {
```
Climbatize, line 54: `isMiniAppAlreadyDownloaded` sounds better to me

MiniApp/Classes/MiniAppDownloader.swift
```diff
@@ -23,27 +23,56 @@ class MiniAppDownloader {
...
+                             self.miniAppStorage.cleanVersions(for: appId, differentFrom: versionId, status: self.miniAppStatus)
+
+                             fallthrough
+                         default:
+                             completionHandler(downloadResult)
+                         }
+                     }
+                 case .failure(let error):
+                     let downloadError = error as NSError
+                     if downloadError.code == NSURLErrorNotConnectedToInternet {
```
corycaywood, line 42: In this case (when it fails to fetch the manifest for a mini app), shouldn't it always try to open the cached Mini App version regardless of the type of error, not just when it's `NSURLErrorNotConnectedToInternet? For example, the device could have temporary connection issues that cause the request to time out. I think this is the approach that Android took.

MiniApp/Classes/MiniAppDownloader.swift
```diff
@@ -23,27 +23,56 @@ class MiniAppDownloader {
...
+            let versionDirectory = FileManager.getMiniAppVersionDirectory(with: appId, and: versionId)
+            var isDirectory: ObjCBool = true
+            if FileManager.default.fileExists(atPath: versionDirectory.path, isDirectory: &isDirectory) {
+                return true
             }
         }
+        return false
+    }
+
+    func getCachedMiniApp(appId: String, downloadError: NSError, completionHandler: @escaping (Result<URL, Error>) -> Void) {
```
corycaywood, line 65: It looks like there isn't a reason that this function needs to be passed a completionHandler. It's not doing anything asynchronous, right? 

I think it would be easier to follow if you make `getCachedMiniApp` return `URL?`. And then in the `download` function, you just need to check if it returned a null value. Then you don't need to pass the `downloadError` or `completionHandler` to `getCachedMiniApp`.

MiniApp/Classes/RealMiniApp.swift
```diff
@@ -48,6 +48,10 @@ internal class RealMiniApp {
                 }
                 self.downloadMiniApp(appInfo: appInfo, completionHandler: completionHandler)
             case .failure(let error):
+                let createError = error as NSError
+                if createError.code == NSURLErrorNotConnectedToInternet {
+                    self.downloadMiniApp(appInfo: appInfo, completionHandler: completionHandler)
```
corycaywood, line 53: What case is this for? In case the device was connected to the internet when you fetched the manifest and then disconnected when you try to download the Mini App? But then the `completionHandler` is being immediately called with failure below, so the result here will never get returned to the user, right? Maybe you meant to add a `return` here?

Is there a way to cover this case within `MiniAppDownloader#download` and have it return the cached Mini App? It seems like functionality which should be part of that class and should be covered by the same functionality as when fetching the manifest fails.

MiniApp/Classes/MiniAppDownloader.swift
corycaywood: Ah sorry, I just checked Android and it's only catching network errors: `UnknownHostException` and `SocketTimeoutException`. So is there an Error type for "timeout" which you can check here? And any other network error types that should be checked?

MiniApp/Classes/MiniAppDownloader.swift
ts-leojoseph-r: Done 👍

MiniApp/Classes/MiniAppDownloader.swift
ts-leojoseph-r: Done 👍 . Added Timeout exception as well

MiniApp/Classes/MiniAppDownloader.swift
ts-leojoseph-r: That's a better approach, updated code now 👍

MiniApp/Classes/RealMiniApp.swift
ts-leojoseph-r: This method will fetch the mini-app information from the platform and compare if it is the latest version and if not we download the latest one. In the `.failure` block (esp for Offline scenario) I was calling download method because the download manager will take care of checking if there is any cached version available.
Now that I have changed a few things in `MiniAppDownloader`, instead of downloading, I am calling the `getCachedMiniApp` method now. Please check it

MiniApp/Classes/MiniAppDownloader.swift
```diff
@@ -23,30 +23,67 @@ class MiniAppDownloader {
...
             }
         }
+        return false
+    }
+
+    /// Check if there is any old version of mini app cached for a given app ID
+    /// - Parameter appId: App ID for which the version directory path to be fetched
+    /// - Returns: URL of the available cached version for a given mini app ID
+    func getCachedMiniApp(appId: String) -> URL? {
+        guard let versionDirectory = FileManager.getMiniAppVersionDirectory(usingAppId: appId) else {
```
corycaywood, line 75: So `FileManager.getMiniAppVersionDirectory` will inference the version id by checking the folder structure under that App ID? But what if there are two version directories? For example, if there were some error where the download failed halfway through downloading a new version. Then `FileManager.getMiniAppVersionDirectory` could return the wrong path?

It seems dangerous to assume that the folder structure will be correct and that it won't have any invalid folders for failed versions. Could you cache the downloaded version in `MiniAppStatus` instead? I think this is what Android is doing.

MiniApp/Classes/RealMiniApp.swift
```diff
@@ -48,6 +48,13 @@ internal class RealMiniApp {
                 }
                 self.downloadMiniApp(appInfo: appInfo, completionHandler: completionHandler, messageInterface: messageInterface)
             case .failure(let error):
+                let createError = error as NSError
```
corycaywood, line 51: This block is duplicate code to `MiniAppDownloader.handleDownloadError`, isn't it? Maybe you could move `handleDownloadError` to this class and then you could handle it in the `.failure` case of `downloadMiniApp`?

ts-leojoseph-r: We can do that, in that case, we need to call self.downloadMiniApp() again in .failure() flow

MiniApp/Classes/RealMiniApp.swift
```diff
@@ -48,6 +48,13 @@ internal class RealMiniApp {
...
+                if Constants.offlineErrorCodeList.contains(createError.code) {
```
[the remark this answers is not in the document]
ts-leojoseph-r: I will move the `offlineErrorCodeList` close to the class but I don't think we can move handleDownloadError to this class. 
And the reason is when **createMiniApp** is called in **RealMiniApp**, it will check the platform for the latest mini-app info, if the request is failed(offline) then we load the cached mini-app, if the request is a success(online) then we continue downloading/cache the version(if it is already downloaded).
So we need to take care of returning the cached version of mini-app in two places esp in offline mode,
1. **RealMiniApp** - while fetching the Mini app info - request failed - serve the cached mini app
2. **MiniAppDownloader** - while downloading in progress - download failed - serve the cached mini app

We can combine these into one, but we need to move the `MiniAppInfoFetcher` implementation into `MiniAppDownloader`

MiniApp/Classes/MiniAppDownloader.swift
ts-leojoseph-r: That's right, I guess I need to make changes in `MiniAppStatus`, in which the way it stores the cache information. Will work on this change, thank you 👍

MiniApp/Classes/RealMiniApp.swift
ts-leojoseph-r: I fixed those changes now

PR #97, opened 2020-06-30
SDK and sample app changes for requesting location permission

ts-leojoseph-r:
# Description
* Updated bridging script that will help the mini-app to communicate with iOS SDK - [JS Bridge Commit](https://github.com/rakutentech/js-miniapp/commit/2b31f6ed23ddf7239d0562134a14188453ef44c7)
* Updated iOS SDK to accept new interface i.e requestPermission and added flow for the same
* Updated sample app with location permission info in the plist and the implementation to show the location permission alert

## Links
- [MINI-915](https://jira.rakuten-it.com/jira/browse/MINI-915)

# Checklist


MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift
```diff
@@ -49,6 +52,36 @@ internal class MiniAppScriptMessageHandler: NSObject, WKScriptMessageHandler {
...
+
+        switch requestPermissionType {
+        case .location:
+            getPermissionResult(requestParam: requestParamValue, callbackId: callbackId)
+        default:
+            executeJavaScriptCallback(responseStatus: .onError, messageId: callbackId, response: MiniAppJavaScriptError.invalidPermissionType.rawValue)
+        }
+    }
+
+    func getPermissionResult(requestParam: String, callbackId: String) {
```
corycaywood, line 70: Shouldn't this function accept `MiniAppPermissionType` instead of `requestParam`?

The `requestParam` here is just being used to create the `MiniAppPermissionType`, but this was already created above in `requestPermission` (as `requestPermissionType `) so you could just pass that to this function.

ts-leojoseph-r: Done 👍

MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift
```diff
@@ -49,6 +52,37 @@ internal class MiniAppScriptMessageHandler: NSObject, WKScriptMessageHandler {
...
+    }
+
+    func getPermissionResult(requestPermissionType: MiniAppPermissionType, callbackId: String) {
+        hostAppMessageDelegate?.requestPermission(permissionType: requestPermissionType) { (result) in
+            switch result {
+            case .success:
+                self.executeJavaScriptCallback(responseStatus: .onSuccess, messageId: callbackId, response: "Allowed")
+            case .failure(let error):
+                if !error.localizedDescription.isEmpty {
+                    self.executeJavaScriptCallback(responseStatus: .onError, messageId: callbackId, response: error.localizedDescription)
```
minh-rakuten, line 78: I see that if users deny permission then it will execute error callback on js bridge.
I guess android should do the same? @corycaywood

corycaywood: Yeah, it's probably better to use the error callback when denied permission.

MiniApp/Classes/JavascriptBridge/MiniAppJavaScriptEnumerations.swift
```diff
@@ -10,4 +11,28 @@ enum JavaScriptExecResult: String {
...
     case unexpectedMessageFormat
+    case invalidPermissionType
+}
+
+public enum MiniAppPermissionType: String {
+    case location
+}
+
+public enum MiniAppPermissionResult: Error {
+    case denied
```
ts-nitish-a-jain, line 22: Should model the same structure as adapted by the OS. Including One-Time Use.

ts-leojoseph-r: Yes, it is same as CLAuthorizationStatus structure

PR #100, opened 2020-07-08
Support Javascript dialog

ts-leojoseph-r:
# Description
Implemented WKUIDelegates to support Javascript dialog - alert, confirm and prompt.

## Links
[MINI-1210](https://jira.rakuten-it.com/jira/browse/MINI-1210)

# Checklist


MiniApp/Classes/Display/RealMiniAppView+WKUIDelegate.swift
```diff
@@ -0,0 +1,48 @@
...
+
+    func webView(_ webView: WKWebView,
+                 runJavaScriptConfirmPanelWithMessage message: String,
+                 initiatedByFrame frame: WKFrameInfo,
+                 completionHandler: @escaping (Bool) -> Void) {
+        let alertController = UIAlertController(title: "Info", message: message, preferredStyle: .alert)
+        alertController.addAction(UIAlertAction(title: "OK", style: .default, handler: { (_) in
+            completionHandler(true)
+        }))
+        alertController.addAction(UIAlertAction(title: "Cancel", style: .cancel, handler: { (_) in
```
corycaywood, line 22: There's also the problem of localization. This text should be in Japanese on a device set to Japanese. But this could be a problem because maybe we would need to add support for other languages in the future. Does iOS have a built-in solution for this?

For example on Android, we can set string resources and then override them with locale-specific resource. Then if an App developer wants to add another language, they can override the string themselves by adding another local-specific resource for the language they want to support. Does iOS have something like this?

Tests/Unit/Helpers.swift
```diff
@@ -245,3 +245,11 @@ func deleteMockMiniApp(appId: String, versionId: String) {
 func deleteStatusPreferences() {
     UserDefaults.standard.removePersistentDomain(forName: "com.rakuten.tech.mobile.miniapp")
 }
+
+func tapAlertButton(title: String, actions: [UIAlertAction]?) {
+    typealias AlertHandler = @convention(block) (UIAlertAction) -> Void
+
+    guard let action = actions?.first(where: {$0.title == title}), let block = action.value(forKey: "handler") else { return }
+    let handler = unsafeBitCast(block as AnyObject, to: AlertHandler.self)
+    handler(action)
```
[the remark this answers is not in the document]
ts-leojoseph-r: Yeah, I don't feel much comfortable either. But it is bit tricky to write tests to AlertController

MiniApp/Classes/Display/RealMiniAppView+WKUIDelegate.swift
```diff
@@ -0,0 +1,48 @@
+import WebKit
+
+extension RealMiniAppView: WKUIDelegate {
+    func webView(_ webView: WKWebView,
+                 runJavaScriptAlertPanelWithMessage message: String,
+                 initiatedByFrame frame: WKFrameInfo,
+                 completionHandler: @escaping () -> Void) {
+        let alertController = UIAlertController(title: "Info", message: message, preferredStyle: .alert)
```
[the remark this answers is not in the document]
ts-leojoseph-r: I see in Android it displays something like this, 
Title - **The page at "<URL>"**
Message - **MESSAGE**

Yeah, we can do similar behavior in iOS. But it would expose the URL loaded in the WebView, is that ok?

MiniApp/Classes/Display/RealMiniAppView+WKUIDelegate.swift
ts-leojoseph-r: Yes, we have Localization in iOS that will help to support different languages, will add it. 👍

MiniApp/Classes/Display/RealMiniAppView+WKUIDelegate.swift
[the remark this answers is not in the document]
corycaywood: Now I'm thinking that we should change Android to use a custom Alert dialog too. We shouldn't be exposing the page URL. I can't think of anything that makes sense to use as the title, except maybe the title of the Mini App. Is there any way to access that from here? It seems that it might make the code a lot more complicated just to inject the Mini App title...

I also think setting the title as just an empty string might be okay. Maybe we should confirm with the product side.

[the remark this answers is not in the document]
ts-nitish-a-jain: Isn't it possible to receive title as an input?

Tests/Unit/Helpers.swift
```diff
@@ -245,3 +245,11 @@ func deleteMockMiniApp(appId: String, versionId: String) {
 func deleteStatusPreferences() {
```
ts-nitish-a-jain, line 245: The file should be refactored for smaller specific units of functionality. What do you think @Climbatize @leorjoseph ?

MiniApp/Classes/Display/RealMiniAppView+WKUIDelegate.swift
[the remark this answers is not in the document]
ts-leojoseph-r: @ts-nitish-a-jain JS dialog doesn't have option to have a title

Tests/Unit/Helpers.swift
ts-leojoseph-r: We use this method in tearDown/afterEach/beforeEach process in which it deletes preferences related to a specific mini-app. As of now, we were not in a need to handle them in different units.

MiniApp/Classes/Display/RealMiniAppView+WKUIDelegate.swift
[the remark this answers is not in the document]
ts-nitish-a-jain: how about making the title to be the display name of the mini app?

Tests/Unit/Helpers.swift
ts-nitish-a-jain: Do you mean to say it is a `MiniAppPrefsRemover`?

ts-leojoseph-r: Yeah, something like that.

MiniApp/Classes/Display/RealMiniAppView+WKUIDelegate.swift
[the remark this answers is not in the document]
ts-leojoseph-r: Cory sent a message to check on the product side if that is ok.

MiniApp/Classes/Display/RealMiniAppView+WKUIDelegate.swift
Climbatize: We already use localizable strings in the Demo app

ts-leojoseph-r: @Climbatize Yes, and this one is for SDK

ts-leojoseph-r: Done

MiniApp/Classes/Display/RealMiniAppView+WKUIDelegate.swift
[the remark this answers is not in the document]
ts-leojoseph-r: Added Mini app displayname as JS dialog title

MiniApp/en.lproj/Localizable.strings
```diff
@@ -1,25 +1,2 @@
...
-
-"wait_message" = "Please wait...";
-"ok" = "OK";
-"cancel" = "Cancel";
-
-"test_mode_publishing" = "Published";
-"test_mode_testing" = "Testing";
-"test_mode_default" = "Default";
+"Ok_title" = "OK";
+"Cancel_title" = "Cancel";
```
corycaywood, line 2: I think the SDK needs to have built-in support for Japanese. Can you add Japanese strings for this?

English | Japanese
-- | --
OK | OK
Cancel | キャンセル

donnie-jp: You should also provide a way for apps to override your localized strings and provide their own. I don't remember the details but I think it just needs the app to define the same string in their `.strings` file but you may need to add support in the SDK code too. The RAuth SDK has this and you can refer to their implementation.

ts-leojoseph-r: @corycaywood Added Jp support

@donnie-jp Sure 👍 , that's a great support to have. I guess we can work on that as a separate task.

PR #103, opened 2020-07-13
Local server implementation - Telegraph

ts-leojoseph-r:
# Description
Integrated [Telegraph](https://github.com/Building42/Telegraph) pod in iOS SDK.

* Mini-apps will be hosted using a local server (https self-signed certificate)

NOTE: Will be removing/add optional dependencies for the existing implementation WKURLSchemeHandler later.

# Checklist


Example/Info.plist
```diff
@@ -45,5 +45,22 @@
 	</array>
 	<key>NSLocationWhenInUseUsageDescription</key>
 	<string>Demo app helps you find more information near you</string>
+	<key>NSAppTransportSecurity</key>
+	<dict>
+		<key>NSExceptionDomains</key>
+		<dict>
+			<key>localhost</key>
```
donnie-jp, line 52: Is there an alternative to using a self-signed cert?
Can we get a real cert from a CA?

Climbatize: yes we can, but would it be unnecessary for our needs ?

ts-leojoseph-r: @donnie-jp , if we can get a real cert from CA that would be awesome. I created it from my Keychain assistant. How do we get one?

corycaywood: But the private key needs to be committed to this repo, right? So the CA would revoke our certificate because the private key has been exposed.

donnie-jp: Sorry, I don't know the details - just flagging it as a possible issue.

corycaywood:
@ts-leojoseph-r @Climbatize @ts-nitish-a-jain I want to start a discussion regarding the local server implementation. Are we sure this is the right approach? The problem that we're trying to solve is that `navigator.geolocation` can't be accessed currently, right? 

To me, implementing a local server seems like a heavy-handed solution. It's a fundamental change to the underlying functionality of our SDK, so it could have unforeseen consequences. Also, could there be any security concerns with this? For example:

- Are other Apps or web browsers on the device (or same network as the device) able to access the local server? Can we ensure that they can't?
- The certificate and private key must be committed to the repo. This certificate won't be used anywhere else, but I'm sure that this will be flagged in security audits.
- This is a more minor concern, but a major component of our SDK would be dependent on a third party library which we don't fully understand it's robustness.

An alternative could be to replace the `navigator.geolocation` APIs with calls to native methods, and this would get around the WebView's secure origin checks. We could use the JS bridge for this purpose:

```
navigator.geolocation.getCurrentLocation = function(success, error, options) {
   MiniAppBridge.exec(
      "getCurrentLocation",
      options,
      success,
      error
}
```

What do you think?

Example/Info.plist
ts-nitish-a-jain: Can we inject the certificate directly into the build process and not include it in the repo?

ts-nitish-a-jain:
> @ts-leojoseph-r @Climbatize @ts-nitish-a-jain I want to start a discussion regarding the local server implementation. Are we sure this is the right approach? The problem that we're trying to solve is that `navigator.geolocation` can't be accessed currently, right?
> 
> To me, implementing a local server seems like a heavy-handed solution. It's a fundamental change to the underlying functionality of our SDK, so it could have unforeseen consequences. Also, could there be any security concerns with this? For example:
> 
> * Are other Apps or web browsers on the device (or same network as the device) able to access the local server? Can we ensure that they can't?
> * The certificate and private key must be committed to the repo. This certificate won't be used anywhere else, but I'm sure that this will be flagged in security audits.
> * This is a more minor concern, but a major component of our SDK would be dependent on a third party library which we don't fully understand it's robustness.
> 
> An alternative could be to replace the `navigator.geolocation` APIs with calls to native methods, and this would get around the WebView's secure origin checks. We could use the JS bridge for this purpose:
> 
> ```
> navigator.geolocation.getCurrentLocation = function(success, error, options) {
>    MiniAppBridge.exec(
>       "getCurrentLocation",
>       options,
>       success,
>       error
> }
> ```
> 
> What do you think?

@corycaywood The two issues in highlight would be: 
- Google Analytics (which is not supported at this moment, critical one) - Business Enabler
- Geolocation (secure origin)
(with the base of a variety of areas which we addressed in past about data separation, redirection, etc.)

There was a comparison from @leorjoseph in past: https://confluence.rakuten-it.com/confluence/display/ONEAPP/Local+server+investigation

@Climbatize @leorjoseph What do you guys think?

corycaywood:
> @corycaywood The two issues in highlight would be:
> 
> * Google Analytics (which is not supported at this moment, critical one) - Business Enabler
> * Geolocation (secure origin)
>   (with the base of a variety of areas which we addressed in past about data separation, redirection, etc.)
> 
> There was a comparison from @leorjoseph in past: https://confluence.rakuten-it.com/confluence/display/ONEAPP/Local+server+investigation
> 
> @Climbatize @leorjoseph What do you guys think?

But my concerns regarding security still stand: I did some testing and found that it's possible to access a currently open mini app from another device on the same network by going to the URL `https://IP_ADDRESS:PORT/MINI_APP_ID`. Is there any way to prevent this?

Regarding the two issues that the local server solves, I think both have alternative solutions:

* Google Analytics: This isn't working when using the custom scheme because of two reasons. By default, google analytics doesn't allow sending from a URL which isn't HTTP/HTTPS, however there is a setting to disable this (`checkProtocolTask`). Also, google analytics stores the client's unique ID as a cookie, but cookies can't be used from a custom scheme URL. However, it's possible to retrieve the client ID and manually cache it in localstorage. See here: https://stackoverflow.com/a/61044459 I tested this on iOS and it's working.
* For `navigator.geolocation`, see my solution above. We can replace this JS function with a call to native code.

donnie-jp:
> What do you think?

I agree with your concerns. Introducing a dependency like that is a big risk in terms of maintenance and knowledge of all the code the SDK relies on. Having a local server running (probably) also opens up another vector of attack.

If you do need a localhost server you maybe need to write your own to replace this dependency in the near future.

You also maybe don't need the private key locally. You could just check that there is a valid certificate on the localhost server without doing the full key verification.

Example/Info.plist
corycaywood: Injecting it at build time might be better.

Or even better, is it possible to generate the certificate at runtime on the device? This is possible on Android, but not sure about iOS...

Climbatize:
> To me, implementing a local server seems like a heavy-handed solution. It's a fundamental change to the underlying functionality of our SDK, so it could have unforeseen consequences. Also, could there be any security concerns with this? For example:
> 
> * Are other Apps or web browsers on the device (or same network as the device) able to access the local server? Can we ensure that they can't?
> * The certificate and private key must be committed to the repo. This certificate won't be used anywhere else, but I'm sure that this will be flagged in security audits.
> * This is a more minor concern, but a major component of our SDK would be dependent on a third party library which we don't fully understand it's robustness.
> 
> An alternative could be to replace the `navigator.geolocation` APIs with calls to native methods, and this would get around the WebView's secure origin checks. We could use the JS bridge for this purpose:
> 
> ```
> navigator.geolocation.getCurrentLocation = function(success, error, options) {
>    MiniAppBridge.exec(
>       "getCurrentLocation",
>       options,
>       success,
>       error
> }
> ```
> 
> What do you think?

With all the arguments you are providing here it is a no go for me too. If we want to go with a local server approach, we have to fork the lib and adapt it to our needs in order to make the local server sandboxed. It is a big task, I think it is achievable, Telegraph if fully coded in swift, but I'm not sure right now. What I am sure is that every time we could have fixed an issue quickly by using a local server we found a bypass, I'm just not sure we always will be able to find one.

ts-leojoseph-r:
I was hoping for these concerns/feedback suggestions at the [investigation findings](https://confluence.rakuten-it.com/confluence/display/ONEAPP/Local+server+investigation) or a heads up before started working on the local server. 
Now that with Cory's suggestions and Brainstorming session with the team by Monday I believe we can come to common grounds.

MiniApp/Classes/Storage/MiniAppLocalServerPreferences.swift
```diff
@@ -0,0 +1,15 @@
...
+
+    init() {
+        self.defaults = UserDefaults(suiteName: "com.rakuten.tech.mobile.miniapp.server")
+    }
+
+    func savePortNumberForMiniApp(appId: String, portNumber: Int) {
+        defaults?.setValue(portNumber, forKey: appId)
+    }
+
+    func getPortNumberForMiniApp(key: String) -> Int {
```
corycaywood, line 12: One thing I haven't been able to understand: How is the port number generated for a Mini App?

We need to make sure that we don't conflict with the default ports on iOS: https://support.apple.com/en-us/HT202944 Also, the max port number is `65535`. Is there anything in place to prevent conflicts or too high of a port number?

ts-leojoseph-r: If we send the port # as `0` and when mini-app is launched for the first time, the local server will start with random available port number (less than `65535`) and the same is stored in defaults. Next time when the mini-app is launched we will use the same port #

corycaywood: Ah, okay nice so it's just generated by the Telegraph library?

But isn't it still possible that it will generate the same port number twice? For example, if you start a Mini App once, then close the Host App, and then start a different Mini App sometime in the future, it could generate (by chance) the same random number for the other Mini App. 

The library doesn't have a cache of which port numbers it has used, right? Should there be some checking added to make sure that the generated port isn't already being used by another Mini App?

Climbatize: By the way, do we really need one port by miniApp ? That virtually means one server by miniApp, right ?

MiniApp/Classes/Display/RealMiniAppView.swift
```diff
@@ -52,12 +58,33 @@ internal class RealMiniAppView: UIView {
...
+        localServer = nil
+        super.removeFromSuperview()
+    }
+
+    func loadWebpage(versionId: String) {
+        guard let serverUrl = localServer?.serverURL().absoluteString else {
+            return
+        }
+
+        let url = URL(string: "\(serverUrl)/\(versionId)/index.html")!
```
corycaywood, line 78: Does this need to have the `versionId` in the path? This means the Mini App URL to be `https://localhost:PORT/VERSION/index.html`, but it seems this causes some Mini Apps to not work (such as the JS SDK Sample App). It's because they are referencing their assets with a preceding slash like `/styles/something.css'.

So this would work if you just remove the `versionId` from the path so that the Mini App URL is `https://localhost:PORT/index.html`.

MiniApp/Classes/Storage/MiniAppLocalServerPreferences.swift
corycaywood: Each one has a different port so that they can have data separation for localStorage, etc. Is there another way to do that without using different ports?

Also, I think that the server is stopped when the Mini App is closed, right?

Climbatize: Oh yes server is stopped during the removeFromSuperView, it was not the case during our very first tests so that's why I wrote this, sorry

ts-leojoseph-r:
Closing this PR, as local server with certificates (https) is no longer required. New PR will be created for local server (without https) just in case if it is required

PR #108, opened 2020-07-20
Enable override localizable strings from Host app

ts-leojoseph-r:
# Description
Enable localizable override from the host app.

# Checklist


ts-leojoseph-r:
@donnie-jp I am not sure how its been done in RAuth (Couldn't find the repo), but have a look at this solution

Tests/Unit/RealMiniAppViewTests.swift
```diff
@@ -24,7 +24,7 @@ class RealMiniAppViewTests: QuickSpec {
             context("when host app info is specified in plist") {
                 it("will add custom string in User agent") {
                     let miniAppView = RealMiniAppView(miniAppId: "miniappid-testing", versionId: "version-id", miniAppTitle: "", hostAppMessageDelegate: mockMessageInterface)
-                    expect(miniAppView.webView.customUserAgent).toEventually(contain("HOSTAPPNAME_AND_VERSION"), timeout: 3)
+                    expect(miniAppView.webView.customUserAgent).toEventually(contain("HOSTAPPNAME_AND_VERSION"), timeout: 6)
```
Climbatize, line 27: I hope we never reach 6 seconds to achieve this test 😅, it happened to you ?

MiniApp/Classes/Display/RealMiniAppView+WKUIDelegate.swift
```diff
@@ -48,7 +48,10 @@ extension RealMiniAppView: WKUIDelegate {
...
         let path = Bundle(for: MiniApp.self).path(forResource: "Localization", ofType: "bundle")!
-        let podBundle = Bundle(path: path) ?? Bundle.main
-        return NSLocalizedString(title, bundle: podBundle, comment: "")
+        guard let podBundle = Bundle(path: path) else {
+            return ""
+        }
+        let defaultValue = NSLocalizedString(title, bundle: podBundle, comment: "")
+        return NSLocalizedString(title, tableName: "Localizable", bundle: Bundle.main, value: defaultValue, comment: "")
     }
 }
```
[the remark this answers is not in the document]
ts-leojoseph-r: Done 👍

Tests/Unit/RealMiniAppViewTests.swift
ts-leojoseph-r: Added 6 secs as a safe timeout. It failed for me like once/twice when I executed them in Local, maybe it takes time to get the value from plist I think.

Climbatize: ok I see, then it's necessary, I think it depends on machine load

PR #112, opened 2020-07-21
change sample UI settings to avoid enabling testing mode

Climbatize:
# Description
We need to release version 1.2 Before backend enables `testing` mode API. SDK is still working with `published` mode, so we just hide mode setting to avoid developers who use demo app to try switching it

## Links
N/A

# Checklist

### DIFF
--- a/Example/SettingsTableViewController.swift
+++ b/Example/SettingsTableViewController.swift
@@ -59,7 +59,7 @@ class SettingsTableViewController: UITableViewController {
         }
     }
     // swiftlint:disable:next todo
-    //FIXME: remove to enable test mode after v1.2 release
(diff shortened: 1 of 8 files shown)


Example/SettingsTableViewController.swift
```diff
@@ -50,14 +50,19 @@ class SettingsTableViewController: UITableViewController {
...
+            return "Test Mode"
+        case 0:
             return "RAS"
         default:
-            return ""
+            return nil
         }
     }
+    // swiftlint:disable:next todo
+    //FIXME: remove to enable test mode after v1.2 release
```
ts-nitish-a-jain, line 62:
```suggestion
    //FIXME: remove to enable test mode after backend API is ready
```

PR #114, opened 2020-07-27
Location - Get current location override from SDK

ts-leojoseph-r:
# Description
`navigator.getCurrentPosition` in any mini-app will return the device's Latitude and Longitude position.

# Checklist


MiniApp/Classes/JavascriptBridge/bridge.js
```diff
@@ -114,6 +114,22 @@ MiniAppBridge.requestPermission = function(permissionType) {
     );
   });
 };
+
+/**
+ Below code will override the navigator.geolocation.getCurrentPosition method
+ */
+navigator.geolocation.getCurrentPosition = function(success, error, options) {
```
corycaywood, line 121: The `navigator.geolocation` shouldn't be modified on Android, so this should be set within `if (isPlatform.iOS())`.

MiniApp/Classes/JavascriptBridge/bridge.js
```diff
@@ -114,6 +114,22 @@ MiniAppBridge.requestPermission = function(permissionType) {
...
+    return MiniAppBridge.exec(
+        "getCurrentPosition",
+        null,
```
corycaywood, line 124: There are some options that can be passed to `getPosition` ([see here](https://developer.mozilla.org/en-US/docs/Web/API/Geolocation/getCurrentPosition)). Does it make sense to support any of them?

Probably `maximumAge` isn't relevant, right? But not sure about `timeout` and `enableHighAccuracy`. Looks like there's a `CLLocationManager#desiredAccuracy`. Can this be changed after initialization?

MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift
```diff
@@ -83,6 +88,11 @@ internal class MiniAppScriptMessageHandler: NSObject, WKScriptMessageHandler {
         }
     }
 
+    func getCurrentPosition(callbackId: String) {
+        let responseString = "{\"coords\":{\"latitude\":\(locationManager?.location?.coordinate.latitude ?? 0),\"longitude\":\(locationManager?.location?.coordinate.longitude ?? 0)}}"
```
corycaywood, line 92: This is missing a few things from [`GeolocationPosition`](https://developer.mozilla.org/en-US/docs/Web/API/GeolocationPosition). 

Does the `CLLocationManager` provide a `timestamp`? If not, maybe we could at least generate one ourselves?

It's also missing some of the [`GeolocationCoordinates`](https://developer.mozilla.org/en-US/docs/Web/API/GeolocationCoordinates) such as `altitude`, `heading`, `speed`, and `altitudeAccuracy`. It looks like it's fine for these properties to be null if it's not possible to retrieve them (some devices don't have these capabilities, so I think that's why they're nullable). However, there's also an `accuracy` property which can't be null. Is it possible to retrieve some value for accuracy?

MiniApp/Classes/JavascriptBridge/bridge.js
ts-leojoseph-r: maximumAge and timeout are irrelevant because the location info will be returned immediately. I have updated the logic to use `enableHighAccuracy` now. If `enableHighAccuracy` is true then CLLocationManager#desiredAccuracy set as `kCLLocationAccuracyBest` and if enableHighAccuracy is false then CLLocationManager#desiredAccuracy set as `kCLLocationAccuracyHundredMeters`

MiniApp/Classes/JavascriptBridge/bridge.js
ts-leojoseph-r: Done

MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift
ts-leojoseph-r: That's right. I guess I will try to add all of them.

MiniApp/Classes/Extensions/MiniApp+Date.swift
```diff
@@ -0,0 +1,6 @@
+extension Date {
+    var currentDate: String {
```
Climbatize, line 2: I would give this a more explicit name, like `epochInMillisecondString` or `epochInMillisecond`

MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift
```diff
@@ -83,6 +88,22 @@ internal class MiniAppScriptMessageHandler: NSObject, WKScriptMessageHandler {
...
+
+    func getLocationInfo() -> String {
+        return  "{\"coords\":{\"latitude\":\(locationManager?.location?.coordinate.latitude ?? 0)" +
+        ",\"longitude\":\(locationManager?.location?.coordinate.longitude ?? 0)" +
+        ", \"altitude\":\(locationManager?.location?.altitude ?? 0)" +
+        ", \"altitudeAccuracy\":\(locationManager?.location?.verticalAccuracy ?? 0)" +
+        ", \"accuracy\":\(locationManager?.location?.horizontalAccuracy ?? 0)" +
+        ", \"speed\":\(locationManager?.location?.speed ?? 0)" +
+        ", \"heading\":\(locationManager?.location?.course ?? 0)" +
+        "}, \"timestamp\":\(Date().currentDate)}"
```
Climbatize, line 104: for more clarity, you can write multiple lines strings with triple quotes `"""` :
```swift
return """
        {
            "coords": {
                "latitude":\(locationManager?.location?.coordinate.latitude ?? 0),
                "longitude":\(locationManager?.location?.coordinate.longitude ?? 0),
                "altitude":\(locationManager?.location?.altitude ?? 0),
                "altitudeAccuracy":\(locationManager?.location?.verticalAccuracy ?? 0),
                "accuracy":\(locationManager?.location?.horizontalAccuracy ?? 0),
                "speed":\(locationManager?.location?.speed ?? 0),
                "heading":\(locationManager?.location?.course ?? 0)
            },
            "timestamp":\(Date().currentDate)
        }
        """
```

MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift
```diff
@@ -92,3 +113,15 @@ internal class MiniAppScriptMessageHandler: NSObject, WKScriptMessageHandler {
         }
     }
 }
+
+class LocationManager: NSObject, CLLocationManagerDelegate {
```
Climbatize, line 117: Why are you implementing `CLLocationManagerDelegate` if you don't use the delegate methods ?

MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift
Climbatize: oh my bad, I guess you just want a single line quote

ts-leojoseph-r: That's right. Multi-line string won't work here

ts-leojoseph-r:
> LGTM, except for the LocationManager object. If your are asking a position you should leave time to to the `CLLocationManagerDelegate` to receive an update before sending it back to Javascript, especially if you're willing to get a high accuracy position. Thus having the delegates implemented and call `self.executeJavaScriptCallback` after the first update should be considered. What do you think about this?
> 
> We can even consider having a mechanism to call javascript every time the location is updated

That's right. I was thinking of adding the same delegate implementation at first. But right now we are supporting only `getCurrentPosition` which will retrieve the user location at once. It doesn't track the user's location. And this can be done using [CLLocationManager#location](https://developer.apple.com/documentation/corelocation/cllocationmanager/1423687-location) will get the location details. 

But if we start supporting the [watchPosition](https://developer.mozilla.org/en-US/docs/Web/API/Geolocation/watchPosition) or [clearWatch](https://developer.mozilla.org/en-US/docs/Web/API/Geolocation/clearWatch) then we need to use the delegates to track the user's location

MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift
ts-leojoseph-r: My bad. Removed it

MiniApp/Classes/Extensions/MiniApp+Date.swift
ts-leojoseph-r: Done

MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift
```diff
@@ -83,6 +88,22 @@ internal class MiniAppScriptMessageHandler: NSObject, WKScriptMessageHandler {
...
+    }
+
+    func getLocationInfo() -> String {
+        return  "{\"coords\":{\"latitude\":\(locationManager?.location?.coordinate.latitude ?? 0)" +
+        ",\"longitude\":\(locationManager?.location?.coordinate.longitude ?? 0)" +
+        ", \"altitude\":\(locationManager?.location?.altitude ?? 0)" +
+        ", \"altitudeAccuracy\":\(locationManager?.location?.verticalAccuracy ?? 0)" +
+        ", \"accuracy\":\(locationManager?.location?.horizontalAccuracy ?? 0)" +
+        ", \"speed\":\(locationManager?.location?.speed ?? 0)" +
+        ", \"heading\":\(locationManager?.location?.course ?? 0)" +
```
corycaywood, line 103: It looks like `altitude`, `altitudeAccuracy`, `speed`, and `heading` are nullable (see https://developer.mozilla.org/en-US/docs/Web/API/GeolocationCoordinates). Is it possible to change the fallback value to `null` instead of `0`?

ts-leojoseph-r: Done 👍

PR #116, opened 2020-07-29
Remove run time config for User Agent

ts-leojoseph-r:
# Description
Remove run time configuration for UA. UA can be configured only at the build time.

# Checklist


corycaywood:
What was the reason behind this change? All of the other configuration options can be set via either .plist or at runtime, so this breaks with the expectations of the SDK user.

If we keep this change, we should probably add some note to the README about runtime config changes being not supported for the user agent.

ts-leojoseph-r:
> What was the reason behind this change? All of the other configuration options can be set via either .plist or at runtime, so this breaks with the expectations of the SDK user.
> 
> If we keep this change, we should probably add some note to the README about runtime config changes being not supported for the user agent.

I checked with @ts-nitish-a-jain and I believe we should allow the Host application to customize the User-Agent string only @ build time. That's why I made this change

MiniApp/Classes/Environment.swift
```diff
@@ -54,7 +52,7 @@ internal class Environment {
     }
 
     var hostAppUserAgentInfo: String {
-        return value(for: customHostAppUserAgentInfo, fallback: .hostAppUserAgentInfo)
+        return value(for: .none, fallback: .hostAppUserAgentInfo)
```
Climbatize, line 55: you could directly return hostAppUserAgentInfo from plist if you never intend to override it.

ts-leojoseph-r:
@corycaywood @Climbatize , Main reason why we don't want to have them enabled at runtime as well is that there is a chance the User Agent may get added more than once(host app can modify the values anytime) in WebView.

corycaywood:
> @corycaywood @Climbatize , Main reason why we don't want to have them enabled at runtime as well is that there is a chance the User Agent may get added more than once(host app can modify the values anytime) in WebView.

Probably this is fine to not support changing it at runtime. But please update the user guide with a note so that it's clear this value is an exception and can't be changed at runtime.

MiniApp/Classes/Environment.swift
ts-leojoseph-r: Done

PR #117, opened 2020-07-29
feat: Verify cached Mini App files using SHA512 hash

corycaywood:
# Description
Opening a draft PR to start a discussion on what I was thinking to solve our cache-poisoning vulnerability. Some of this probably needs re-working since I don't know iOS/Swift very well and most of the security stuff was copied from other places (stackoverflow). The `KeyStore` class is almost exactly copied from the [Remote Config SDK](https://github.com/rakutentech/ios-remote-config/blob/master/RRemoteConfig/KeyStore.swift), so we might want to think about moving that class to the [SDK Utils Repo](https://github.com/rakutentech/ios-sdkutils).

After downloading a Mini App, an SHA256 hash will be generated for all of the files. Next, a hash is generated for the sum of these hashes and then this final hash value is stored in the device KeyChain. If the files that are cached on the device are modified, then the hash value in the keychain will no longer match, so the Mini App will be deleted and re-downloaded.

Unfortunately, this will slow down the initial launch time for the Mini App a little, but I tried some tests and found that the hash is calculated pretty fast. iOS seems to do some internal optimizations related to this, so the first time that the hash is calculated it takes longer but will be faster after that.

Small App (JS SDK Sample App): 20ms to create hash first time (7ms after that)
Very Large App (27MB and 1000+ files):  ~300ms hash time (70ms after that) 

## Links
MINI-919

# Checklist


ts-leojoseph-r:
Definitely is good to have. So it will store/update Hash for every time the mini-app is downloaded. I don't think the launch time impact isn't that bad.

PR #119, opened 2020-08-04
Feature/build secrets

Climbatize:
# Description
fix to keep compatibility with existing regular URL 

# Checklist


configure-secrets.sh
```diff
@@ -17,7 +17,9 @@ do
   fi
 done
 
-RMA_API_ENDPOINT_SECRET=${RMA_API_ENDPOINT:=https:/\$()/www.example.com}
+ENDPOINT=${RMA_API_ENDPOINT:=https://www.example.com}
+
+RMA_API_ENDPOINT_SECRET="${ENDPOINT/\/\///\$()/}"
```
corycaywood, line 22: Nice, too bad it's so complicated looking with so many slashes 😣  Can you add a comment? Something like: `# Replace "//" with "/$()/" because xcconfig files treat "//" as a comment`.

Also you should remove the lines below about `echo "In order to have the // in https://,...`.

Climbatize: I created a function, much cleaner

Climbatize: I left the comment because it is added to the XCConfig file that a developer would want to modify manually for some reason

PR #127, opened 2020-08-14
User profile settings - Name & Image from the library

ts-leojoseph-r:
# Description
Added User data settings for the sample app.
User should be able to,
* **Name** - Add/Modify the display name
* **Profile Picture** - Add/Change the Profile picture from the Library.
* **Contacts** - Displaying dummy contacts - each contact associated with random UUID.

<table><tr><td>
    
</td>
<td>
</td>
<td>
     
</td>
</tr>
</table>

<table><tr><td>
    
</td>
<td>
</td>
<td>
     
</td>
</tr>
</table>

## Links
[MINI-954](https://jira.rakuten-it.com/jira/browse/MINI-954)

# Checklist


Example/ContactsListSettingsTableViewController.swift
```diff
@@ -0,0 +1,24 @@
...
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        self.tableView.separatorStyle = .singleLine
+        self.tableView.tableFooterView = UIView(frame: .zero)
+    }
+
+    override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        let cell = tableView.dequeueReusableCell(withIdentifier: "Cell", for: indexPath) as UITableViewCell? ?? UITableViewCell()
+        cell.textLabel?.text = "User Contact - " + String(indexPath.row + 1)
```
corycaywood, line 16: Can we make the label be the actual UUID instead? The UUID is the only thing that will be passed to the Mini App, so this label isn't useful to the user.

Example/ContactsListSettingsTableViewController.swift
```diff
@@ -0,0 +1,24 @@
+import UIKit
+import MiniApp
+
+class ContactsListSettingsTableViewController: UITableViewController {
```
corycaywood, line 4: I think the contacts should be editable too, right? We can automatically populate the list with random UUIDs and then allow them to add/delete items.

corycaywood (commented):
Looks really good, just a few comments.

And probably we need to wait until we find out more about the profile image before merging this.

Example/UserSettingsTableViewController.swift
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    func retrieveProfileSettings(key: String = "ProfileImage") -> UIImage? {
+        if let userProfile = UserDefaults.standard.data(forKey: key) {
+            let userProfileData = try? PropertyListDecoder().decode(UserProfileModel.self, from: userProfile)
+            self.displayNameTextField.text = userProfileData?.displayName
+            guard let imageData = userProfileData?.profileImageData else {
+                return nil
+            }
+            return UIImage(data: imageData)
```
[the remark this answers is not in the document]
corycaywood: I just got an update and it's looking like the image will be provided as a URL. Is it possible to provide a URI to the JavaScript side for the image that was chosen by the picker? I think we specifically designed it so the JS side can't access anything else on the file system, right?

So maybe this will need to be changed to accept an image URL instead...

Example/ContactsListSettingsTableViewController.swift
ts-leojoseph-r: Right now there are like 10 dummy contacts added to the list (associated with random UUID's). Right now we don't allow the user to edit/add the contacts. Do you want to allow the user to add/delete the items from the list ?? Deleting the item from the list wouldn't be a big functionality though, but adding a new item to the list needs some little effort

Example/ContactsListSettingsTableViewController.swift
ts-leojoseph-r: Yes, we can. I thought displaying UUID would confuse the user. I will do it 👍

Example/ContactsListSettingsTableViewController.swift
ts-leojoseph-r: Done. Right now you can swipe to delete the contact from the list and added a + button in the right top corner. When + button is tapped, it will automatically add a new UUID list in the list.

Example/ContactsListSettingsTableViewController.swift
ts-leojoseph-r: Done

Example/UserSettingsTableViewController.swift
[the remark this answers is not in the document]
ts-leojoseph-r: We can change it to URI, that URI will be coming from `tmp` path. If that is ok ??

Example/ContactsListSettingsTableViewController.swift
corycaywood: Sorry, I wasn't very clear on this. The IDs that are passed by the Host App won't be UUIDs, they will be some sort of custom ID scheme. QA needs to be able to test a variety of different inputs in the contact list.

So on this screen, when the user taps the "+" button, it should allow them to input the ID they want to add.

Example/UserSettingsTableViewController.swift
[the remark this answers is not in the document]
corycaywood: @ts-leojoseph-r But will the Mini App be able to access a URI from `tmp` path? I thought that a Mini App can't access a file outside of its own files.

I just thought of another idea: we could convert the image to base64 and pass it as a data URI. See here: https://stackoverflow.com/a/25516291 . What do you think?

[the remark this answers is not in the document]
ts-leojoseph-r: Ah, I think this would work for sure. I will update the code

Example/ContactsListSettingsTableViewController.swift
ts-leojoseph-r: Understood 👍 , I will change that too. But the list which displays this UUID is fine right? Also would like to know if these items in the list (random UUIDs and the value added by the QA) should persist?

Example/UserSettingsTableViewController.swift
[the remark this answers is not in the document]
ts-leojoseph-r: Done

corycaywood:
It's looking good! But I'm seeing an issue with it when I test: If I save a profile image and user name and then return to the User Settings screen again, it doesn't show the saved image and name. Probably it should show the saved values, right?

Example/ContactsListSettingsTableViewController.swift
```diff
@@ -0,0 +1,90 @@
...
+
+    override func tableView(_ tableView: UITableView, canEditRowAt indexPath: IndexPath) -> Bool {
+        return true
+    }
+
+    override func tableView(_ tableView: UITableView, commit editingStyle: UITableViewCell.EditingStyle, forRowAt indexPath: IndexPath) {
+        if editingStyle == .delete {
+            if userContactList.indices.contains(indexPath.row) {
+                userContactList.remove(at: indexPath.row)
+                self.tableView.reloadData()
```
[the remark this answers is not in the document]
ts-leojoseph-r: Done

Example/SettingsTableViewController.swift
```diff
@@ -40,6 +46,7 @@ class SettingsTableViewController: UITableViewController {
         self.textFieldSubKey.delegate = self
         addTapGesture()
         addBuildVersionLabel()
+        self.tableView.separatorStyle = .singleLine
```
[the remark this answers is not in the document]
ts-leojoseph-r: For some reason, it wasn't working even after I changed in Storyboard. So I added here as well

ts-leojoseph-r:
> It's looking good! But I'm seeing an issue with it when I test: If I save a profile image and user name and then return to the User Settings screen again, it doesn't show the saved image and name. Probably it should show the saved values, right?

Ahhh, I am really sorry. I forgot to update the key while saving the details in plist. Now it is working now after the changes, please check it.

corycaywood (approved):
Looks good 👍

PR #128, opened 2020-08-24
fix: (demo app) Callback was not triggered after requesting location permission

corycaywood:
# Description
Set CLLocationManager.delegate to self in ViewController. The CLLocationManagerDelegate was defined, but it wasn't getting set to the locationManager, so this means that the callback wasn't called after the user accepted/denied the location permission

## Links
MINI-1558

# Checklist


Climbatize (approved):
nice catch

ts-leojoseph-r (approved):
Looks good to me :+1:
I tried the same without delegate while implementing and checked the same in sample location mini-app, I was able to get the location.

PR #129, opened 2020-08-26
Feat/custom permissions interface

ts-leojoseph-r:
# Description
Custom Permissions implementation on the SDK side. Features include the following,

* Implementation to allow the mini-app to request the predefined custom permissions
* **Public interface API's** for the host app to store and retrieve the custom permissions for a given mini-app ID
* Mini-app permissions are cached in the SDK
* Host app UI to request the custom permissions from the user (The same UI can be reused for [MINI-958](https://jira.rakuten-it.com/jira/browse/MINI-958) with some minor tweaks)

### Sample Request from Mini app
```
[
    {"name":"rakuten.miniapp.USER_NAME", "description": "Mini app needed User name for registration"},
    {"name":"rakuten.miniapp.PROFILE_PHOTO", "description": "Mini app needed Profile pic for registration"},
    {"name":"rakuten.miniapp.PHONE_NUMER", "description": "Mini app needed Phone number for offers"}
]
```
### Sample Response

```
{
  permissions: [
    {"name":"rakuten.miniapp.USER_NAME", "isGranted": "ALLOWED"},
    {"name":"rakuten.miniapp.PROFILE_PHOTO", "isGranted": "DENIED"},
    {"name":"rakuten.miniapp.PHONE_NUMER", "isGranted": "PERMISSION_NOT_AVAILABLE"}
  ]
}
```
## Links
[MINI-953](https://jira.rakuten-it.com/jira/browse/MINI-953)
[MINI-952](https://jira.rakuten-it.com/jira/browse/MINI-952)

# Checklist


Example/ViewController+MiniAppMessageProtocol.swift
```diff
@@ -25,6 +26,17 @@ extension ViewController: MiniAppMessageProtocol, CLLocationManagerDelegate {
         }
     }
 
+    func requestCustomPermissions(permissions: [MASDKCustomPermissionModel], completionHandler: @escaping (Result<[MASDKCustomPermissionModel], Error>) -> Void) {
+        if let viewController = UIStoryboard(name: "Main", bundle: nil).instantiateViewController(withIdentifier: "CustomPermissionsTableViewController") as? CustomPermissionsTableViewController {
+            viewController.customPermissionHandlerObj = completionHandler
+            viewController.permissionsRequestList = permissions
+            viewController.miniAppTitle = self.currentMiniAppTitle ?? "MiniApp"
+            let navController = UINavigationController(rootViewController: viewController)
+            navController.modalPresentationStyle = .fullScreen
```
Climbatize, line 35: 👌

Example/ViewController.swift
```diff
@@ -54,6 +56,7 @@ class ViewController: UIViewController {
             displayController?.miniAppDisplay = miniAppDisplay
             self.currentMiniAppInfo = nil
             self.currentMiniAppView = nil
+            self.displayController = displayController
```
Climbatize, line 59: we could use the `self.presentedViewController as? DisplayNavigationController` when needed instead of creating a `displayController` variable. Do you think it could be relevant here?

MiniApp/Classes/Storage/MiniAppStatus.swift
```diff
@@ -1,10 +1,12 @@
 class MiniAppStatus {
     private let defaults: UserDefaults?
     private let miniAppInfoDefaults: UserDefaults?
+    private let miniAppCustomPermissionsDefaults: UserDefaults?
```
Climbatize, line 4: should we use keychain for this? UserDefaults can easily be modified and allowing permissions is sensible. We can use a lib that makes the keychain behave like a UserDefault or just implement a helper class for it

Climbatize (approved):
Very good, as usually. I think we could make an attempt to have some default UI implemented into the SDK so the host app could use custom permissions like native permissions.

MiniApp/Classes/Storage/MiniAppStatus.swift
```diff
@@ -44,4 +46,26 @@ class MiniAppStatus {
...
+            }
+            if let data = try? PropertyListEncoder().encode(permissionsToStore) {
+                self.miniAppCustomPermissionsDefaults?.set(data, forKey: id)
+            }
+        }
+    }
+
+    func getCustomPermissions(forMiniApp id: String) -> [MASDKCustomPermissionModel]? {
+        if let data = self.miniAppCustomPermissionsDefaults?.data(forKey: id) {
+            let miniAppInfo = try? PropertyListDecoder().decode([MASDKCustomPermissionModel].self, from: data)
```
corycaywood, line 66: I think we need to return the full list of supported permissions here. So even if a permission has never been requested, it should be returned here as `DENIED`.

MiniApp/Classes/Storage/MiniAppStatus.swift
```diff
@@ -44,4 +46,26 @@ class MiniAppStatus {
         }
         return nil
     }
+
+    func setCustomPermissions(forMiniApp id: String, permissionList: [MASDKCustomPermissionModel]) {
```
corycaywood, line 50: I think we shouldn't cache the `description`. This is a dynamic value which could change in the Mini App (i.e. between different Mini App versions, different device language settings, etc.). I think we don't have a requirement for showing this description on the permission revoking screen.

MiniApp/Classes/JavascriptBridge/MiniAppJavaScriptPermissionManager.swift
```diff
@@ -0,0 +1,156 @@
...
+    func checkCustomPermissionsRequestStatusInCache(miniAppPermissionRequestModelList: [MASDKCustomPermissionModel], callbackId: String) {
+        /// storedPermissionsList - List of Permissions that is already stored in the cache
+        if let storedPermissionsList = self.miniAppStatus.getCustomPermissions(forMiniApp: self.miniAppId) {
+            /// filteredRequestModelList - List of Permissions that user hasn't responded yet.
+            let userNotRespondedRequestList = miniAppPermissionRequestModelList.filter {
+                !storedPermissionsList.contains($0)
+            }
+            userRespondedRequestList = storedPermissionsList.filter { miniAppPermissionRequestModelList.contains($0)
+            }
+            /// If the User hasn't responded to more than one permission that is requested, request Host app to request Custom Permissions
```
corycaywood, line 28: I think this comment isn't needed: code is self explanatory.

MiniApp/Classes/JavascriptBridge/MiniAppJavaScriptPermissionManager.swift
```diff
@@ -0,0 +1,156 @@
...
+            if userNotRespondedRequestList.count > 0 {
+                requestHostApp(customPermissionRequestList: userNotRespondedRequestList, callbackId: callbackId)
+            } else {
+                /// If the user has already responded to all the permissions that is requested filteredRequestModelList
```
corycaywood, line 32: Same here, I think this comment isn't needed. It seems like the kind of comment that could quickly become outdated and not get updated, so it's better to let the code speak for itself.

MiniApp/Classes/JavascriptBridge/MiniAppJavaScriptPermissionManager.swift
```diff
@@ -0,0 +1,156 @@
...
+                /// will be 0 and the below code will filter the list of permissions(requested) from the stored permissions
+                /// and the same will be returned as cached response.
+                self.sendCachedSuccessResponse(result: userRespondedRequestList, callbackId: callbackId)
+            }
+        } else {
+            /// if User hasn't responded to any Custom permissions before
```
corycaywood, line 38: Same here, I think this comment isn't needed. 

Also, see my comment above about filtering only `ALLOWED` permissions. I think that will simplify this logic because then we don't have a distinction between `Permissions which have been allowed`, `Permissions which have been denied`, and `Permissions which have not been requested`.

There should only be `Permissions which have been allowed` and `Permissions which are denied (need to be requested)`.

corycaywood (commented):
It's looking good 👍 

The main issue is the handling of "Denied" permissions. You have implemented it with the concept of "Accepted", "Denied", and "Not yet requested". 

However, I think we should have only the concept of "Accepted" and "Denied". So if a user chooses "Deny" for a certain permission, they will still prompted about that permission in the future if it's requested again. This will make some of the things in the SDK simpler, it allows the Host App and Mini App to have more flexibility (the Mini App can still request a Denied permission again), and it allows us to provide all of the supported permissions when returning `getCustomPermissions`, rather than just those which have been requested before.

Let me know if you have any questions on this.

Example/CustomPermissionsTableViewController.swift
```diff
@@ -0,0 +1,107 @@
...
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        self.tableView.tableFooterView = UIView(frame: .zero)
+        self.addBarButtonItems()
+        self.addFooterInfo()
+    }
+
+    func addBarButtonItems() {
+        let allowBarButtonItem = UIBarButtonItem(title: "Allow", style: .done, target: self, action: #selector(allowPermissions))
```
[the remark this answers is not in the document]
ts-leojoseph-r: Almost forgot 👍 . Will add it 👍

MiniApp/Classes/JavascriptBridge/MiniAppJavaScriptPermissionManager.swift
```diff
@@ -0,0 +1,156 @@
+extension MiniAppScriptMessageHandler {
+    func requestCustomPermissions(requestParam: RequestParameters?, callbackId: String) {
+        cachedUnknownCustomPermissionRequest.removeAll()
+        userRespondedRequestList.removeAll()
+        guard let requestParamValue = requestParam?.customPermissions else {
```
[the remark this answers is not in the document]
Climbatize: I thought the same

MiniApp/Classes/Storage/MiniAppStatus.swift
khairul-alam-licon: In Android, I am using `PERMISSION_NOT_AVAILABLE` as default to return when user never asked for a custom permission.

MiniApp/Classes/Storage/MiniAppStatus.swift
donnie-jp: Permissions is something important/sensitive enough that it shouldn't be stored in userdefaults or anywhere in plaintext. Keychain sounds like a good option.

Example/AppDelegate.swift
```diff
@@ -9,5 +9,19 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+      return topViewController(controller: tabController.selectedViewController)
+    }
+    if let navController = controller as? UINavigationController {
+      return topViewController(controller: navController.visibleViewController)
+    }
+    if let presented = controller?.presentedViewController {
+      return topViewController(controller: presented)
+    }
+    return controller
+  }
```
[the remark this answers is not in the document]
ts-leojoseph-r: Done

Example/CustomPermissionsTableViewController.swift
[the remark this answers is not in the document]
ts-leojoseph-r: Done

Example/CustomPermissionsTableViewController.swift
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+
+    func getSwitchView(tagValue: Int) -> UISwitch {
+        let switchView = UISwitch(frame: .zero)
+        switchView.setOn(true, animated: true)
+        switchView.tag = tagValue
+        switchView.addTarget(self, action: #selector(permissionValueChanged(_:)), for: .valueChanged)
+        return switchView
+    }
```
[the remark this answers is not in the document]
ts-leojoseph-r: I thought it would be great to have native behavior(like in iOS Settings cell-associated toggle buttons) where the cell selection is disabled and only the toggle is enabled. Adding the toggle in the storyboard, well I thought of using the default cells but I will definitely think about adding a custom cells 👍

Example/CustomPermissionsTableViewController.swift
```diff
@@ -0,0 +1,107 @@
...
+    func addFooterInfo() {
+        let footerLabel = UILabel()
+        footerLabel.text = "  \(miniAppTitle) wants to access the above permissions. Choose your preference accordingly.\n\n  You can also manage these permissions later in the Miniapp settings"
+        footerLabel.textAlignment = .center
+        footerLabel.textColor = .gray
+        footerLabel.numberOfLines = 0
+        tableView.addSubview(footerLabel)
+        footerLabel.translatesAutoresizingMaskIntoConstraints = false
+        footerLabel.bottomAnchor.constraint(equalTo: tableView.safeAreaLayoutGuide.bottomAnchor).isActive = true
+        footerLabel.widthAnchor.constraint(equalTo: tableView.safeAreaLayoutGuide.widthAnchor).isActive = true
```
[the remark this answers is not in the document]
ts-leojoseph-r: While adding a footer in UITableViewController via storyboard, it sticks next to the tableview last cell, but I wanted to have the footer at the bottom of the page. That's why I went with this approach

Example/ViewController.swift
ts-leojoseph-r: Oh sorry. This object is not used anymore. My bad, I removed this one now.

MiniApp/Classes/Storage/MiniAppStatus.swift
ts-leojoseph-r: Understood @Climbatize , @donnie-jp . Will make the changes 👍 . Thank you.

MiniApp/Classes/JavascriptBridge/MiniAppJavaScriptPermissionManager.swift
ts-leojoseph-r: This was a Big PR and I felt like adding a few comments that would help the reviewers. I was planning to remove them before merging to the master

MiniApp/Classes/Storage/MiniAppStatus.swift
ts-leojoseph-r: Based on our recent discussion, I will be removing the description from the caching.

MiniApp/Classes/Storage/MiniAppStatus.swift
ts-leojoseph-r: @corycaywood , is that true? I was not aware of that requirement(couldn't find it in JIRA too). I will make the changes accordingly 👍

ts-leojoseph-r: @khairul-alam-licon we would never return `PERMISSION_NOT_AVAILABLE` when `getCustomPermissions` as it contains the only a list of supported permissions and its granted status.

MiniApp/Classes/JavascriptBridge/MiniAppJavaScriptPermissionManager.swift
ts-leojoseph-r: > However, I think we should have only the concept of "Accepted" and "Denied". So if a user chooses "Deny" for a certain permission, they can still be prompted about that permission in the future if it's requested again. This will make some of the things in the SDK simpler, it allows the Host App and Mini App to have more flexibility (the Mini App can still request a Denied permission again), and it allows us to provide all of the supported permissions when returning `getCustomPermissions`, rather than just those which have been requested before.

Thanks for the comment cory. Yes, I implemented something like iOS behavior, where permission which is denied by the user cannot be requested again. I will make changes to as per the comments and to retrieve supported permissions when `getCustomPermissions`  is called

MiniApp/Classes/JavascriptBridge/MiniAppJavaScriptPermissionManager.swift
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    func checkCustomPermissionsRequestStatusInCache(miniAppPermissionRequestModelList: [MASDKCustomPermissionModel], callbackId: String) {
+        /// storedPermissionsList - List of Permissions that is already stored in the cache
+        if let storedPermissionsList = self.miniAppStatus.getCustomPermissions(forMiniApp: self.miniAppId) {
+            /// filteredRequestModelList - List of Permissions that user hasn't responded yet.
+            let userNotRespondedRequestList = miniAppPermissionRequestModelList.filter {
+                !storedPermissionsList.contains($0)
+            }
+            userRespondedRequestList = storedPermissionsList.filter { miniAppPermissionRequestModelList.contains($0)
```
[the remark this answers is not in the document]
ts-leojoseph-r: Thy will be done

Example/CustomPermissionsTableViewController.swift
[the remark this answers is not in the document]
Climbatize: oh understood, then we should have a `UIViewController` implementing `UITableViewDelegate` and `UITableViewDataSource` with a simple table view and cell prototype. Simpler to maintain. I really think that less code for UI is better, at least as long as we are using `UIStoryboard`s

Example/CustomPermissionsTableViewController.swift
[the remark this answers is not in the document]
Climbatize: This is just a demo app, so you don't have to break all the good work just because of this advice. In this case the behavior you are attempting to achieve is maybe better with a XIB containing the permission view and a `UIStackView` inside a `UIScrollView` to deal with permissions display. But honestly I would also have gone with the UITableView 👍

MiniApp/Classes/Storage/MiniAppStatus.swift
khairul-alam-licon: Got it. Thanks.

MiniApp/Classes/Storage/MiniAppKeyChain.swift
```diff
@@ -0,0 +1,124 @@
...
+                    rawValue: $0.rawValue)!,
+                isPermissionGranted: .denied,
+                permissionRequestDescription: ""
+            ))
+        }
+        write(keys: [id: supportedPermissionList])
+        return supportedPermissionList
+    }
+
+    private func write(keys: KeysDictionary) {
```
corycaywood, line 63: Is there any way you could use the [KeyStore from iOS SDK Utils
](https://github.com/rakutentech/ios-sdkutils/blob/master/RSDKUtils/KeyStore/KeyStore.swift) to simplify this class? It seems like the writing and reading functionality is similar. We're already using the `KeyStore` class for [cache verification](https://github.com/rakutentech/ios-miniapp/blob/master/MiniApp/Classes/Storage/MiniAppCacheVerifier.swift).

Sorry, I'm not very familiar with iOS KeyChain, so I'm not sure if this makes sense to use it or not.

Example/ViewController+MiniAppMessageProtocol.swift
```diff
@@ -25,6 +26,17 @@ extension ViewController: MiniAppMessageProtocol, CLLocationManagerDelegate {
         }
     }
 
+    func requestCustomPermissions(permissions: [MASDKCustomPermissionModel], completionHandler: @escaping (Result<[MASDKCustomPermissionModel], Error>) -> Void) {
+        if let viewController = UIStoryboard(name: "Main", bundle: nil).instantiateViewController(withIdentifier: "CustomPermissionsTableViewController") as? CustomPermissionsTableViewController {
```
corycaywood, line 30: I think this should first check if all of the permissions are already granted. If they are already granted then it should immediately invoke the callback to the JS bridge. This is because we will probably suggest that the Mini Apps call `requestCustomPermission` every time they use any of the user data features (because the user could revoke the permissions at any time from the Host App settings). What do you think?

Example/ViewController+MiniAppMessageProtocol.swift
```diff
@@ -25,6 +26,17 @@ extension ViewController: MiniAppMessageProtocol, CLLocationManagerDelegate {
...
+            viewController.customPermissionHandlerObj = completionHandler
+            viewController.permissionsRequestList = permissions
```
corycaywood, line 32: One of the requirements is `If perms have already been granted, the user shouldn't need to grant them again.`

So I think this means it that if the user requests a list of permissions which contains both granted and not yet granted permissions, then the user should not be shown the already accepted permissions again. So here you should filter out the permissions which have already been granted before passing to the ViewController so that it only displays toggles for the permissions which have not yet been granted. Do you think that makes sense?

MiniApp/Classes/MiniApp.swift
```diff
@@ -51,6 +51,26 @@ public class MiniApp: NSObject {
...
+    public func setCustomPermissions(forMiniApp appId: String, permissionList: [MASDKCustomPermissionModel]) {
+        if !appId.isEmpty {
+            return realMiniApp.storeCustomPermissions(forMiniApp: appId, permissionList: permissionList)
+        }
+    }
+
+    /// Get the list of supported Custom permissions and its status for a given Mini app ID
+    /// - Parameter appId: Mini AppId String value
+    /// - Returns: List of MASDKCustomPermissionModel that contains the details of every custom permission type, status and description.
+    public func getCustomPermissions(forMiniApp appId: String) -> [MASDKCustomPermissionModel]? {
```
corycaywood, line 67: I think that the return value shouldn't be nullable. This should always return the full list of supported permissions. If they have never requested one of the permissions, then that permission should be returned as `DENIED`. This is so the Host App can have a consistent permissions screen for every Mini App. Also a user may wish to grant a particular permission to a Mini App which the Mini App hasn't requested yet. However, the downside is that it would be possible to grant a permission which the Mini App doesn't actually use.

Do you think that makes sense?

corycaywood (commented):
It's getting really close! A few more comments below.

Also could you please update the documentation?
- [ ] Update the userguide with instructions for implementing custom permissions
- [ ] Update changelog

MiniApp/Classes/Storage/MiniAppKeyChain.swift
ts-leojoseph-r: That's right. I have used the most of the Keystore wrapper from SDK Utils, but our SDK needs a different dictionary [String: [MASDKCustomPermissionModel]] to be stored, so I thought I could write a similar interface in our SDK.

Example/ViewController+MiniAppMessageProtocol.swift
ts-leojoseph-r: We are already checking in the SDK if the user has already granted permissions (before requesting the host app) and if yes, then we are sending the cached response back to the Mini app. [Check here](https://github.com/rakutentech/ios-miniapp/pull/129/files#diff-ee09dac8c84e5da89672c0c27e4f6401R37)

Example/ViewController+MiniAppMessageProtocol.swift
ts-leojoseph-r: We have already taken care of it in SDK side, [User already responded list](https://github.com/rakutentech/ios-miniapp/pull/129/files#diff-ee09dac8c84e5da89672c0c27e4f6401R26) and [User not responded list](https://github.com/rakutentech/ios-miniapp/pull/129/files#diff-ee09dac8c84e5da89672c0c27e4f6401R30)

MiniApp/Classes/MiniApp.swift
ts-leojoseph-r: Agreed 👍 , return shouldn't be nullable. Will make changes

MiniApp/Classes/Storage/MiniAppKeyChain.swift
corycaywood: Ah, okay it's slightly different. And it looks like the `KeyStore` class doesn't currently let you retrieve keys. Maybe in the future we can have a technical improvement to make `KeyStore` more generic so that it can be used in this class.

ts-leojoseph-r: That's the idea. I was even thinking to improve this class (Adding more interface)to take care of the Hashing technique as well.

ts-leojoseph-r:
> * [x]  Update the userguide with instructions for implementing custom permissions
> * [x]  Update changelog
Done

corycaywood (approved):
Looks good 👍

Example/CustomPermissionsTableViewController.swift
```diff
@@ -0,0 +1,107 @@
...
+
+    override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        let cell = tableView.dequeueReusableCell(withIdentifier: "cell", for: indexPath)
+        let permissionModel: MASDKCustomPermissionModel?
+        if permissionsRequestList?.indices.contains(indexPath.row) ?? false {
+            permissionModel = permissionsRequestList?[indexPath.row]
+            cell.textLabel?.text = permissionModel?.permissionName.title
+            cell.detailTextLabel?.text = permissionModel?.permissionDescription
+            cell.detailTextLabel?.numberOfLines = 0
+            cell.accessoryView = getSwitchView(tagValue: indexPath.row)
```
Climbatize, line 87: nice approach

MiniApp/Classes/Storage/MiniAppKeyChain.swift
```diff
@@ -0,0 +1,126 @@
...
+        guard let permissionList = allKeys[keyId] as [MASDKCustomPermissionModel]? else {
+            return setDefaultPermissionsInKeyChain(forMiniApp: keyId)
+        }
+        return permissionList
+    }
+
+    func storeCustomPermissions(permissions: [MASDKCustomPermissionModel], forMiniApp keyId: String) {
+        var keysDic = keys()
+        var cachedPermissions = self.getCustomPermissions(forMiniApp: keyId)
+        _ = permissions.map { (permissionModel: MASDKCustomPermissionModel) -> MASDKCustomPermissionModel in
```
Climbatize, line 27: did you need to assign it to the `_` ?

Climbatize (approved):
You went very far with the keychain implementation, I just suggested using an existing wrapper library :) Nice job indeed

MiniApp/Classes/Storage/MiniAppKeyChain.swift
ts-leojoseph-r: Yeah, I just want to map permissions and make changes in cachedPermissions. So the result is not needed

ts-leojoseph-r:
> You went very far with the keychain implementation, I just suggested using an existing wrapper library :) Nice job indeed

I was hoping to make use of the existing interfaces in SDKUtils but later I changed my mind and I felt that we definitely need a wrapper in the future. Thank you!!

PR #130, opened 2020-09-01
Updated documentation for public class/variables and some formatting

ts-leojoseph-r:
# Description
* Added some documentation for public APIs
* Refactoring

# Checklist

### DIFF
--- a/MiniApp/Classes/Extensions/MiniApp+FileManager.swift
+++ b/MiniApp/Classes/Extensions/MiniApp+FileManager.swift
@@ -35,8 +35,14 @@ extension FileManager {
     class func getMiniAppVersionDirectory(usingAppId id: String) -> URL? {
         do {
             let miniAppDirectory = getMiniAppDirectory(with: id)
-            let versionDirectory = try FileManager.default.contentsOfDirectory(at: miniAppDirectory, includingPropertiesForKeys: [URLResourceKey.isDirectoryKey], options: [.skipsHiddenFiles])[0]
-            return versionDirectory
+            var isDir: ObjCBool = true
+            if FileManager.default.fileExists(
+                atPath: miniAppDirectory.path,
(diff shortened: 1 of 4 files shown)


MiniApp/Classes/Extensions/MiniApp+FileManager.swift
```diff
@@ -35,8 +35,14 @@ extension FileManager {
     class func getMiniAppVersionDirectory(usingAppId id: String) -> URL? {
         do {
             let miniAppDirectory = getMiniAppDirectory(with: id)
-            let versionDirectory = try FileManager.default.contentsOfDirectory(at: miniAppDirectory, includingPropertiesForKeys: [URLResourceKey.isDirectoryKey], options: [.skipsHiddenFiles])[0]
-            return versionDirectory
+            var isDir: ObjCBool = true
+            if FileManager.default.fileExists(
+                atPath: miniAppDirectory.path,
+                isDirectory: &isDir) {
```
corycaywood, line 41: Why does this need to be passed as a variable instead of just directly passing `true`?

corycaywood (approved):
Looks good! 👍

MiniApp/Classes/Extensions/MiniApp+FileManager.swift
ts-leojoseph-r: We need to pass a [UnsafeMutablePointer](https://developer.apple.com/documentation/foundation/filemanager/1410277-fileexists) - ObjC boolean here. I wish I would know a better way to do here, but sort of followed the old way.

Climbatize: yeah, unfortunately this is a C++ legacy from ObjC.

Climbatize (approved):
LGTM

PR #132, opened 2020-09-02
Updated all errors to conform to MiniAppErrorProtocol

ts-leojoseph-r:
# Description

* Updated SDK to conform to `MiniAppErrorProtocol`, that helps to follow a structure like below that is sent back to the Mini app.
    ```json
   ERROR_NAME: ERROR_DESCRIPTION
    ```
for Eg.,

```
    invalidPermissionType: Permission type that is requested is invalid
```

* Updated MiniAppMessageProtocol - completion handlers to return results with error that conforms to MiniAppErrorProtocol

# Checklist


corycaywood (changes requested):
This is a good idea! But we can't change the error response for already existing functionality because this is a breaking change for a mini app. They are expecting a string error response for `requestLocationPermission` and `getUniqueId`. It's fine to have an error object for new functionality like `requestCustomPermissions`, but please revert to a string response for already existing functions.

Or maybe we could keep these changes and update the `bridge.js` code? So for example, on the already existing functions, the `bridge.js` code would parse out the error message so that it can return just a string as expected by the Mini App. This might make it easier to migrate to using an error object format in the future.

ts-leojoseph-r:
> Or maybe we could keep these changes and update the `bridge.js` code? So for example, on the already existing functions, the `bridge.js` code would parse out the error message so that it can return just a string as expected by the Mini App. This might make it easier to migrate to using an error object format in the future.

I guess making changes in bridge.js would be a better idea. I really felt bad that I didn't come up with some better error structure in the first place, but I want to push this change now before its too late, especially before these custom permissions go live.

ts-leojoseph-r:
@corycaywood 
I didn't want to make changes in `bridge.js` for now as the code seems clean and generic for both Android & iOS. Instead, I added an extra method in iOS SDK to send the error response like before for `requestLocationPermission` and `getUniqueId`. Let me know if those changes are fine.

Climbatize (approved):
I like it, much cleaner. Good job

MiniApp/Classes/Models/MiniAppJavascriptErrorInfo.swift
```diff
@@ -27,8 +25,8 @@ public enum MiniAppCustomPermissionError: String, Codable {
     }
 }
 
-func getMiniAppCustomPermissionError(customPermissionError: MiniAppCustomPermissionError) -> String {
-    return getErrorJsonResponse(error: MiniAppError(error: MiniAppErrorDetail(title: customPermissionError, description: customPermissionError.description)))
+func getMiniAppErrorMessage<T: MiniAppErrorProtocol>(_ error: T) -> String {
+    return getErrorJsonResponse(error: MiniAppError(error: MiniAppErrorDetail(title: error.name, description: error.message)))
```
corycaywood, line 29: This isn't correct, is it? It should be returning the same format as a [Javascript Error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) object: `{name: 'name', description: 'description'}`.

MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift
```diff
@@ -35,7 +35,7 @@ internal class MiniAppScriptMessageHandler: NSObject, WKScriptMessageHandler {
     func handleBridgeMessage(responseJson: MiniAppJavaScriptMessageInfo?) {
         guard let actionCommand = responseJson?.action, !actionCommand.isEmpty,
             let callbackId = responseJson?.id, !callbackId.isEmpty else {
-            executeJavaScriptCallback(responseStatus: .onError, messageId: "", response: MiniAppJavaScriptError.unexpectedMessageFormat.rawValue)
+                executeJavaScriptCallback(responseStatus: .onError, messageId: "", response: getMiniAppErrorMessage(MiniAppJavaScriptError.unexpectedMessageFormat))
```
corycaywood, line 38: This means that the error is still going to be received as a string by the JS SDK, right? 

I think that we should be directly passing an [error object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error). So maybe we need to update the JS Bridge code (i.e. here: https://github.com/rakutentech/js-miniapp/blob/master/js-miniapp-bridge/src/common-bridge.ts#L191) so that it parses the string and creates an Error object? Something like this (in typescript):
```
class MiniAppError extends Error {
    constructor(errorResponse: string) {
        super();
        const error = JSON.parse(errorResponse);
        this.name = error.name;
        this.message = error.message;
    }    
}
```

corycaywood (commented):
This is looking better, but see my comments below.

Basically, I think that the `.catch` for the `Promise` should be receiving an `Error` object instead of a string (which the user would have to parse themselves). What do you think?

MiniApp/Classes/Models/MiniAppJavascriptErrorInfo.swift
ts-leojoseph-r: Ah, I removed the `error` now, it will be `{name: 'name', description: 'description'}` now.

MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift
ts-leojoseph-r: Got it 👍 , will create PR in JS SDK

MiniApp/Classes/Models/MiniAppJavascriptErrorInfo.swift
ts-leojoseph-r: This method is updated to return String as `ERROR_NAME: ERROR_DESCRIPTION`

MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift
ts-leojoseph-r: I guess this change is no longer needed.

corycaywood (approved):
Looks good 👍

corycaywood (approved):
👍

PR #134, opened 2020-09-03
tests: Disable test which is failing intermittently on CI

corycaywood:
# Description
We will need to re-enable this test when we find a better way of testing it.

# Checklist


Climbatize (approved):
I'll check how I can improve this test

PR #137, opened 2020-09-10
External link management

Climbatize:
# Description
Adds the possibility to manage external link opening on host app and send a feedback URL to the SDK. A default `SFSafariViewController` embedded into the SDK is used if no delegate is provided by the host app.

## Links
mini-1034

# Checklist


USERGUIDE.md
```diff
@@ -207,20 +207,59 @@ MiniApp iOS SDK provides a fully customizable way to implement a navigation inte
...
+The `externalLinkResponseHandler` closure allows you to give a feedback as an URL to the SDK, for example when the controller is closed or when a custom scheme link is tapped. Here is an example following previous example:
+
+```swift
+extension ExternalWebViewController: WKNavigationDelegate {
+    func webView(_ webView: WKWebView, didFinish navigation: WKNavigation!) {
+        self.currentURL = self.webView.url
+    }
+
+    public func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction, decisionHandler: @escaping (WKNavigationActionPolicy) -> Void) {
+        if let urlWebview = navigationAction.request.url, urlWebview.scheme?.starts(with: "mscheme") ?? false {
```
corycaywood, line 252: I think this functionality should be contained in the SDK - the Host App shouldn't have to know anything about the `mscheme` custom scheme. Also, I think this should only redirect if the custom scheme matches the Mini App ID. i.e. if the custom scheme is redirecting to a different Mini App than the one which launched the WebView, then it should be treated as an invalid URL.

On Android SDK, we created a `MiniAppExternalUrlLoader` class which they can use for this functionality ([see here](https://github.com/rakutentech/android-miniapp/blob/master/miniapp/USERGUIDE.md#automatic-check-in-webview-which-belongs-to-separated-activity)). Is there a way on iOS to have some `shouldOverideUrlLoading` function which they can use?

corycaywood (commented):
Looking good 👍  I left a few comments.

USERGUIDE.md
```diff
@@ -207,20 +207,59 @@ MiniApp iOS SDK provides a fully customizable way to implement a navigation inte
...
                     customNavigationView: mCustomView)
 
 MiniApp.shared(with: Config.getCurrent(), navigationSettings: navConfig).info(miniAppId: miniAppID) { (result) in
 ...
 }
 ```
 
+#### Opening external links
+
+By default MiniApp iOS SDK will open external links into a separate modal controller when tapped. `MiniAppNavigationDelegate` implements a method that allows to override this behaviour and provide your own external links management. Here is an example of implementation:
```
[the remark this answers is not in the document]
Climbatize: This is the case @corycaywood 👍

MiniApp/Classes/Display/RealMiniAppView.swift
```diff
@@ -99,23 +99,27 @@ internal class RealMiniAppView: UIView {
...
+        if let scheme = requestURL.scheme {
+            let schemeType = MiniAppSupportedSchemes(rawValue: requestURL.scheme!)
+            switch schemeType {
+            case .tel:
+                UIApplication.shared.open(requestURL, options: [:], completionHandler: nil)
+            default:
+                if scheme.starts(with: Constants.miniAppSchemePrefix) {
+                    return decisionHandler(.allow)
+                } else {
+                    if navigationAction.navigationType == .linkActivated {
```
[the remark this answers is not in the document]
Climbatize: ok got it

USERGUIDE.md
Climbatize: I'll make an attempt with this in mind

ts-leojoseph-r (commented):
Looks goods to me except for the comments which @corycaywood mentioned. Also can we avoid big refactoring like this in the feature branch unless if it is really needed for the feature. I understand this refactoring is much awaited one though.

MiniApp/Controllers/Extensions/ViewController+MiniApp.swift
```diff
@@ -1,9 +1,27 @@
...
+    func miniAppNavigation(shouldOpen url: URL, with externalLinkResponseHandler: @escaping (URL) -> Void) {
+        if let viewController = UIStoryboard(name: "Main", bundle: nil).instantiateViewController(withIdentifier: "ExternalWebviewController") as? ExternalWebViewController {
+            viewController.currentURL = url
+            viewController.miniAppExternalNavigationHandler = externalLinkResponseHandler
+            self.presentedViewController?.present(viewController, animated: true)
+        }
+    }
+
+    func miniAppNavigation(canUse actions: [MiniAppNavigationAction]) {
+
```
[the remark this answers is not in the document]
Climbatize: you're totally right, I added comments on public delegates

MiniApp/Classes/Display/RealMiniAppView.swift
```diff
@@ -99,23 +100,29 @@ internal class RealMiniAppView: UIView {
...
-        case .tel:
-            UIApplication.shared.open(requestURL, options: [:], completionHandler: nil)
-            decisionHandler(.cancel)
-        default:
-            decisionHandler(.allow)
+    func validateScheme(requestURL: URL, navigationAction: WKNavigationAction, decisionHandler: @escaping (WKNavigationActionPolicy) -> Void) {
+        if let scheme = requestURL.scheme {
+            let schemeType = MiniAppSupportedSchemes(rawValue: scheme)
+            switch schemeType {
+            case .about:
```
corycaywood, line 107: What is the `about` scheme used for?

corycaywood (approved):
Looks good! 👍

MiniApp/Classes/Display/RealMiniAppView.swift
Climbatize: the about scheme is the scheme used for js alert dialogs, for example. This one brought me some challenge to find out

corycaywood: Ah, okay. Maybe you should add a comment so we don't delete it by accident?

ts-leojoseph-r (approved):
Looks good to me :+1:

PR #141, opened 2020-09-15
JS Bridge updates for Custom Permission support

corycaywood:
# Description
- build: Update JS Bridge submodule to latest version
  - This fixes an issue where the response was being returned to the user as as string rather than an object
- refactor: (Custom permissions) Rename 'isGranted' param to 'status'
  - This param doesn't represent a boolean - it represents a status of allowed, denied, or not available so it should be named accordingly.

# Checklist


ts-leojoseph-r (approved):
Looks good to me :+1:

Climbatize (approved):
Good ! I always prefer enum than booleans for statuses, we never know

ts-leojoseph-r:
> Good ! I always prefer enum than booleans for statuses, we never know

@Climbatize was it boolean before 🤷

khairul-alam-licon (approved):
👍

PR #142, opened 2020-09-21
Feat/revoke custom permissions

ts-leojoseph-r:
# Description
* Display Permissions button in User Settings screen
* Display List of Downloaded apps in Mini Apps list screen (If any keys found in Keychain other than downloaded apps, the same will be deleted from the Keychain)
    * List will also display description text that denotes the list of permissions the user has agreed for that mini-app
* Allow Users to Revoke/Change the permissions in Permissions Settings screen.

# Screenshots
<table><tr><td>
    
</td>
<td>
</td>
<td>
     
</td>
</tr>
</table>
<table><tr><td>
    
</td>
<td>
</td>
<td>
     
</td>
</tr>
</table>

## Links
[MINI-958](https://jira.rakuten-it.com/jira/browse/MINI-958)

# Checklist


Example/Images.xcassets/switch.imageset/Contents.json
```diff
@@ -0,0 +1,26 @@
+{
```
corycaywood, line 1: This icon isn't showing up for me when I tested (iPhone 8 Plus (13.5)). It's strange because the other icons are showing up okay.

MiniApp/Classes/MiniApp.swift
```diff
@@ -71,6 +71,14 @@ public class MiniApp: NSObject {
         return []
     }
 
+    /// Gets the list of downloaded Mini apps info and associated custom permissions status
+    /// - Returns:Dictionary of MiniAppInfo and respective custom permissions info
```
corycaywood, line 75: Is this return format convenient to use on iOS? Basically `MiniAppInfo` is the key and `MASDKCustomPermissionModel` is the value, right?

From the Android perspective, this would be a little difficult to use. On Android, we would probably use `Pair` which allows two values to be tied together in a single object. Like this: `Array<Pair<MiniAppInfo, MASDKCustomPermissionModel>>` and then the resulting usage looks something like this:

```
val pairs = Array(
  miniAppInfo1 to permissions1,
  miniAppInfo2 to permissions2,
 // ...
)

pairs.forEach {
  val info = it.first
  val permissions = it.second
  // use data...
}
```

Does iOS have something like `Pair`? Or maybe we could create a custom model for this like `MiniAppInfoPermissionsPair`? What do you think?

MiniApp/Classes/MiniApp.swift
```diff
@@ -71,6 +71,14 @@ public class MiniApp: NSObject {
...
+    public func listDownloadedWithCustomPermissions() -> [[MiniAppInfo: [MASDKCustomPermissionModel]]] {
```
corycaywood, line 76: Could you also add this to the user guide?

corycaywood (commented):
Looks good! Just left a few comments

corycaywood:
@leorjoseph Looks like you accidentally reverted the JS bridge to an old version ([see here](https://github.com/rakutentech/ios-miniapp/pull/142/files#diff-1e63e016bc7dda489f3229486210fa59). 

The submodule works like a normal git repo, so you can update it to the correct version by going to the `js-miniapp` folder and pulling the latest changes from the [js-bridge-ios branch](https://github.com/rakutentech/js-miniapp/tree/js-bridge-ios). Then go back to the normal ios project root folder and commit the changes.

MiniApp/Classes/MiniApp.swift
ts-leojoseph-r: Done. Added in User guide and Changelog

Example/Images.xcassets/switch.imageset/Contents.json
ts-leojoseph-r: That's strange. I tested in like 4 devices including iPhone 8 Plus 13.4.1, I am able to see the images.

ts-leojoseph-r:
> The submodule works like a normal git repo, so you can update it to the correct version by going to the `js-miniapp` folder and pulling the latest changes from the [js-bridge-ios branch](https://github.com/rakutentech/js-miniapp/tree/js-bridge-ios). Then go back to the normal ios project root folder and commit the changes.

Oops, sorry my bad. Yes, that was by accident. Updated the code now.

MiniApp/Classes/MiniApp.swift
ts-leojoseph-r: Yes, in iOS we have [Tuple](https://docs.swift.org/swift-book/ReferenceManual/Types.html). I guess we can go with Tuple/Pair approach instead of custom model. Looks much more cleaner.

ts-leojoseph-r: Done

Example/Images.xcassets/switch.imageset/Contents.json
corycaywood: Strange, now the opposite is happening to me: The icon for settings is showing but the other two icons aren't 😆  Maybe it's something weird about my build environment....

corycaywood (approved):
Looks good 👍

Climbatize (approved):
LGTM too 👍

PR #144, opened 2020-09-28
Fix/Mini-1694, 1771

rleojoseph:
# Description
* Fix - Remove Trailing whitespace when the user taps on Save button
* Fix - No need to validate User name in User settings - _"No field is explicitly required."_ 
* Delete existing Custom permissions for the miniapp from the Keychain when it is downloaded for the first time
* Refactor - Show Image gallery when the user taps on the Image in the User Settings screen

## Links
* [MINI - 1694](https://jira.rakuten-it.com/jira/browse/MINI-1694) 
* [MINI - 1771](https://jira.rakuten-it.com/jira/browse/MINI-1771) 

# Checklist


corycaywood (approved):
👍 👍

Example/Controllers/UserSettingsTableViewController.swift
```diff
@@ -30,6 +31,16 @@ class UserSettingsTableViewController: UITableViewController, UIImagePickerContr
         self.userProfileImage = profileImage
     }
 
+    func addTapGestureForImage() {
+        let tapGestureRecognizer = UITapGestureRecognizer(target: self, action: #selector(imageTapped(tapGestureRecognizer:)))
+        imageView.isUserInteractionEnabled = true
+        imageView.addGestureRecognizer(tapGestureRecognizer)
```
Climbatize, line 37: why not using a button? Less code is better to me ^^

Climbatize (approved):
LGTM, we always have opposite visions on UI, but that's for the best :)

Example/Controllers/UserSettingsTableViewController.swift
ts-leojoseph-r: Lol. 👍 . Added in refactor To-Do's.

PR #145, opened 2020-09-29
Feature/mini 1789: Allow inline video playback

Climbatize:
# Description
Currently mini apps can't display videos inline when <video playsinline> is explicitly coded in the html script.
We set the inline configuration to true into the SDK to allow this behavior

## Links
[MINI-1789](https://jira.rakuten-it.com/jira/browse/MINI-1789)

# Checklist

Screen Recording 2020-09-29 at 17 17 03


MiniApp/Classes/Display/MiniAppWebView.swift
```diff
@@ -10,6 +10,8 @@ internal class MiniAppWebView: WKWebView {
         config.preferences.javaScriptEnabled = true
         config.preferences.setValue(true, forKey: "allowFileAccessFromFileURLs")
         config.setURLSchemeHandler(URLSchemeHandler(versionId: versionId), forURLScheme: schemeName)
+        config.allowsInlineMediaPlayback = true
+        config.allowsPictureInPictureMediaPlayback = true
```
corycaywood, line 14: Does the mini app need to add anything to use picture-in-picture? Or just adding `playsinline` will enable both?

corycaywood (commented):
Looks good 👍  But looks like you reverted the `js-miniapp` sub-module to an old commit, so please either change this back or update it to the latest commit.

Also, could you add this to the change-log? I think we won't include this in v2.2.0, so maybe create a new section for v2.3.0?

MiniApp/Classes/Display/MiniAppWebView.swift
Climbatize: No, this is the default behavior for all videos into a basic web view. In fact it seems like adding `playsinline` disables PiP. I just added this to the config to consolidate the default value, so we have a consistent behavior in Mini App SDK if the default value of this parameter was ever changed by Apple in a future update.

CHANGELOG.md
```diff
@@ -26,6 +26,10 @@
 - **Feature:** Added sample implementation for Sharing the message from Mini app
 - **Feature:** Added sample app implementation to revoke/manage the custom permissions for the list of downloaded mini apps
 
+### 2.x
+
+***SDK***
+-**Fixed:** support of `playsinline` instruction of  `Video` html tag
```
ts-leojoseph-r, line 32: I think this has to be moved to the top `2.x.0`. That has our next release updates

ts-leojoseph-r (approved):
Looks good to me :+1: Just one comment. Please check

CHANGELOG.md
corycaywood: Sorry, this should be move to a separate release version above the other `2.x.0` - this change won't be included in the `2.1.0` release.

Climbatize: men! I thought I was safe with 2.x as it's just not even a version number XD

corycaywood (approved):
👍

PR #149, opened 2020-10-03
Retrieve User name and profile photo

ts-leojoseph-r:
# Description
* Added support for retrieving User Name(`getUserName()`) and Profile Photo(`getProfilePhoto()`) from Mini App via Javascript bridge
* Added support for Host app to implement `MiniAppUserInfoProtocol` for `getUserName` and `getProfilePhoto` interfaces
* Added sample app implementation for the same
* Added swiftlint script under Build Phases -> Run Script that helps to identify Swiftlint errors at build time.

## Links
*  [MINI-1684](https://jira.rakuten-it.com/jira/browse/MINI-1684)
* [MINI-1686](https://jira.rakuten-it.com/jira/browse/MINI-1686)

# Checklist
- [ ] I have read the [contributing guidelines](../CONTRIBUTING.md).


MiniApp/Classes/Protocols/MiniAppMessageProtocol.swift
```diff
@@ -19,6 +19,18 @@ public protocol MiniAppMessageProtocol: class {
     func shareContent(info: MiniAppShareContent, completionHandler: @escaping (Result<MASDKProtocolResponse, Error>) -> Void)
 }
 
+/**
+ Public Protocol that will be used by the Mini App to communicate
+ with the Native implementation for User profile related retrieval
+*/
+public protocol MiniAppUserInfoProtocol: class {
```
Climbatize, line 26: this should be renamed as MiniAppUserInfoDelegate to conform to the delegation pattern best practice used in Swift

MiniApp/Classes/Protocols/MiniAppMessageProtocol.swift
```diff
@@ -19,6 +19,18 @@ public protocol MiniAppMessageProtocol: class {
...
+    /// Interface that is used to retrieve the user name from the User Profile
+    func getUserName() -> String
```
Climbatize, line 28: should we return a `String?` instead?

MiniApp/Classes/Protocols/MiniAppMessageProtocol.swift
```diff
@@ -19,6 +19,18 @@ public protocol MiniAppMessageProtocol: class {
...
+
+    /// Interface that is used to retrieve the Image URI
+    func getProfilePhoto() -> String
```
Climbatize, line 31: should we return a `String?` instead?

Climbatize (commented):
Seems good, tell me what you think of my reviews

MiniApp/Classes/MiniApp.swift
```diff
@@ -47,8 +47,12 @@ public class MiniApp: NSObject {
     ///                         to interact with View component of mini app.
     ///         -   Error: Error details if Mini App View creating is failed
     ///   - messageInterface: Protocol implemented by the user that helps to communicate between Mini App and native application
-    public func create(appId: String, completionHandler: @escaping (Result<MiniAppDisplayProtocol, Error>) -> Void, messageInterface: MiniAppMessageProtocol) {
-        return realMiniApp.createMiniApp(appId: appId, completionHandler: completionHandler, messageInterface: messageInterface)
+    public func create(appId: String,
+                       completionHandler: @escaping (Result<MiniAppDisplayProtocol, Error>) -> Void,
+                       messageInterface: MiniAppMessageProtocol,
+                       userInfoProtocol: MiniAppUserInfoProtocol? = nil
```
corycaywood, line 53: I'm not really sure about this - We will probably need to be adding even more protocols in the future, so this `create` constructor could end up with way too many parameters if we add it here every time there's a new one 😅 

On the Android side, we added a `setUserInfoHandler` method to `MiniAppMessagePrototocol`. Would this be possible and make sense on iOS?

MiniApp/Classes/Protocols/MiniAppMessageProtocol.swift
corycaywood, line 28: I think I misunderstood something about how these protocols are working on iOS. When we add new functions to this protocol in the future, then the Host App will be required to implement the new functions or else their build will fail, right?

I had assumed this wasn't the case because in `RealMiniApp` we are implementing default handlers for the functions. However, these default handlers only apply if the Host App doesn't provide `MiniAppMessageBridgeProtocol` or `MiniAppUserInfoProtocol` at all, correct?

This is a problem because it means we will introduce breaking changes every time we want to add new methods to the protocol. So we need to think of how we can add new functions without the Host App being required to implement them. 

On the Android side, we switched from using `abstract` functions to just normal `open` functions which have either a default implementation or throw an error when the Host App hasn't overridden them. Is something similar possible for iOS? You lose the benefit of requiring a certain implementation, but maybe that's okay because the message interfaces are intended to be optional anyways.

Any ideas @leorjoseph @Climbatize ?

CHANGELOG.md
```diff
@@ -2,9 +2,19 @@
...
-***SDK***
+**SDK**
+
+- **Fixed:** support of `playsinline` instruction of  `Video` html tag
+
+- **Feature:** Added separate public `MiniAppUserInfoProtocol` to communicate with the host app for User profile related retrieval.
+- **Feature:** Added interfaces such as [getUserName](USERGUIDE.md#cuser-profile-details-username) and [getProfilePhoto](USERGUIDE.md#cuser-profile-details-profilephoto) in `MiniAppUserInfoProtocol` to retrieve username and profile photo respectively. Interfaces will be called only if user has agreed to the respective custom permission. i.e `rakuten.miniapp.user.USER_NAME` for [getUserName](USERGUIDE.md#cuser-profile-details-username) and `rakuten.miniapp.user.PROFILE_PHOTO` for [getProfilePhoto](USERGUIDE.md#cuser-profile-details-profilephoto)
+- **Feature:** Added support for Javascript bridge interface for User Info detail retrieval from Mini app.
+`getUserName()` and `geProfilePhoto()`
+- **Feature:** Updated public interface with `MiniAppUserInfoProtocol` support `MiniApp#create(appId:completionHandler:messageInterface:userInfoProtocol)`
```
corycaywood, line 13: I think this seems too detailed for the change log. All four of these items are the same feature: providing user info to the mini app. I think this should be simplified and combined into a single change log entry which just mentions that we've added a new feature for providing user info, so they can implement `MiniAppUserInfoProtocol`, `getUsername`, and `getProfilePhoto`.

CHANGELOG.md
```diff
@@ -2,9 +2,19 @@
 
 ### 2.x.x (In progress)
 
-***SDK***
+**SDK**
+
+- **Fixed:** support of `playsinline` instruction of  `Video` html tag
+
```
corycaywood, line 8: You don't need an extra space here - this causes separate lists to be rendered so there's an extra gap between the items.

corycaywood (commented):
It's looking good 👍  But there are some issues with the public interface we should think about.

MiniApp/Classes/Protocols/MiniAppMessageProtocol.swift
Climbatize: Yeah unfortunately sift does not allow optional methods. There are two ways to solve this, either by annotating the protocol with @objc and use exclusively Objc compliant objects into it, this seems the case here. This way you can make functions optional and have a way to override their implementation if they are not implemented. Either by making subprotocols to keeps swift compliance, you can take a look at my current #151 PR to have an example.

Another approach would be to subclass UIViewController, UItableViewConreoller, UICollectionViewControler, etc. as MiniAppViewController, MiniAppViewController, MiniAppTableViewController, etc. to implement a base behavior, and then ask host developers to rely on these classes instead of base UIKit classes. This is the smartest approach to have abstracts methods, but it can quickly become a nightmare for developers of existing apps who just want to add support for a MiniApp and already have subclasses of UIKit components.

ts-leojoseph-r: I guess Protocol extensions can overcome this problem. Host app can conform to the existing protocols(which we have already) and we can create a variable that conforms to multiple/other protocols. I will share the example project in Slack and we can take the decision forward on this.

CHANGELOG.md
ts-leojoseph-r: Done

MiniApp/Classes/Protocols/MiniAppMessageProtocol.swift
ts-leojoseph-r: Done

MiniApp/Classes/Protocols/MiniAppMessageProtocol.swift
ts-leojoseph-r: Done

MiniApp/Classes/MiniApp.swift
ts-leojoseph-r: We have used Protocol Composition and Protocol extensions to overcome this issue. Now the Host app doesn't need to conform to any new protocols and also Protocol extensions helps the SDK to add a default implementations for the same, which means, the newly added protocols will be optional for the Host app.

MiniApp/Classes/Protocols/MiniAppMessageProtocol.swift
ts-leojoseph-r: @Climbatize , We can make use of Protocol extensions to avoid these situations. I have updated the code, have a look.

MiniApp/Classes/RealMiniApp.swift
```diff
@@ -166,3 +179,14 @@ extension RealMiniApp: MiniAppMessageProtocol {
         completionHandler(.failure(error as Error))
     }
 }
+
+extension RealMiniApp: MiniAppUserInfoProtocol {
+
+    func getUserName() -> String {
+        return ""
```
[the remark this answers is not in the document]
ts-leojoseph-r: ShareContent uses a completion handler that enables us to pass down the error, but `getUserName` and `getProfilePhoto` are returning `string`. We are returning `nil` as of now and `MiniAppScriptMessageHandler` will take care of throwing `internalError`.

Example/Controllers/Extensions/ViewController+MiniAppUserInfoProtocol.swift
```diff
@@ -0,0 +1,18 @@
+import MiniApp
+
+extension ViewController {
```
corycaywood, line 3: Shouldn't this be `extension ViewController: MiniAppUserInfoDelegate`?

MiniApp/Classes/Protocols/Protocol Extensions/MiniAppUserInfoProtocolExtension.swift
```diff
@@ -0,0 +1,9 @@
+public extension MiniAppUserInfoDelegate {
```
corycaywood, line 1: I wonder if it makes more sense to define the extension in the same file as the protocol? Especially because it's a pretty short protocol right now. This would keep them closer together so it's obvious that there's a default implementation of the protocol.

corycaywood (approved):
Nice! This looks good 👍 

Let's wait to get @Climbatize opinion on the protocol extensions.

MiniApp/Classes/Protocols/Protocol Extensions/MiniAppUserInfoProtocolExtension.swift
Climbatize: I was just thinking the same thing

Example/Controllers/Extensions/ViewController+MiniAppUserInfoProtocol.swift
ts-leojoseph-r: We have used Protocol composition. Since `ViewController` has already conformed to `MiniAppMessageProtocol`, we don't need the host app to conform to `MiniAppUserInfoDelegate`

MiniApp/Classes/Protocols/Protocol Extensions/MiniAppUserInfoProtocolExtension.swift
ts-leojoseph-r: Done

MiniApp/Classes/Protocols/MiniAppMessageProtocol.swift
ts-leojoseph-r: Done

MiniApp/Classes/MiniApp.swift
```diff
@@ -47,7 +47,9 @@ public class MiniApp: NSObject {
     ///                         to interact with View component of mini app.
     ///         -   Error: Error details if Mini App View creating is failed
     ///   - messageInterface: Protocol implemented by the user that helps to communicate between Mini App and native application
-    public func create(appId: String, completionHandler: @escaping (Result<MiniAppDisplayProtocol, Error>) -> Void, messageInterface: MiniAppMessageProtocol) {
+    public func create(appId: String,
+                       completionHandler: @escaping (Result<MiniAppDisplayProtocol, Error>) -> Void,
+                       messageInterface: MiniAppMessageProtocol) {
```
Climbatize, line 52: this line is a good example that we need to unify our class naming. For swift developers it's more familiar to deal with delegates when using interfaces or protocols. I know it's 3 different names for the same thing, but this is making use of the delegate pattern, so would it make more sense to use something like `messageDelegate: MiniAppMessageDelegate`. What do you thing about it @corycaywood and @leorjoseph ?

Climbatize (approved):
Protocol extension seems a good way to deal with default implementations. Good job.
However, we are facing a small naming issue with our delegates, my intuition leads me to think we should rename all the calls and classes from Interfaces and Protocols to Delegates.

CHANGELOG.md
```diff
@@ -2,9 +2,17 @@
 
 ### 2.x.x (In progress)
 
-***SDK***
+**SDK**
+
+- **Fixed:** support of `playsinline` instruction of  `Video` html tag
+- **Feature:** Added separate public `MiniAppUserInfoDelegate` to communicate with the host app to [getUserName](USERGUIDE.md#cuser-profile-details-username) and [getProfilePhoto](USERGUIDE.md#cuser-profile-details-profilephoto). Interfaces will be called only if user has agreed to the respective custom permission. i.e `rakuten.miniapp.user.USER_NAME` for [getUserName](USERGUIDE.md#cuser-profile-details-username) and `rakuten.miniapp.user.PROFILE_PHOTO` for [getProfilePhoto](USERGUIDE.md#cuser-profile-details-profilephoto)
+- **Feature:** Added support for Javascript bridge interface for User Info detail retrieval from Mini app.
+`getUserName()` and `geProfilePhoto()`
```
khairul-alam-licon, line 10: I think, we wanted to write "getProfilePhoto()"

khairul-alam-licon (approved):
👍

MiniApp/Classes/MiniApp.swift
ts-leojoseph-r: I completely agree with this. We should definitely unify them, but I wonder that this could break the existing host app which has integrated them already. We should handle it a little carefully. but haven’t you done this change in the other PR of yours??

BTW, you are tagging the wrong person here, please use **@ts-leojoseph-r**  😆

PR #151, opened 2020-10-08
Migrate sharing controller from demo app to SDK

Climbatize:
# Description
Allows mini apps developers to have a default sharing controller if they don't want to code one.
I also renamed `MiniAppMessageProtocol` to `MiniAppMessageDelegate` to keep usual naming in Swift, with backward compatibility.

## Links
MINI-1696

# Checklist


corycaywood:
Looking good, but there are related changes in #149 . So let's wait to merge that first and see how it affects this one.

MiniApp/Classes/Protocols/MiniAppMessageDelegate.swift
```diff
@@ -1,8 +1,11 @@
+@available(*, deprecated, message: "protocol renamed to MiniAppMessageDelegate")
+public typealias MiniAppMessageProtocol = MiniAppMessageDelegate
+
 /**
 Public Protocol that will be used by the Mini App to communicate
  with the Native implementation
 */
-public protocol MiniAppMessageProtocol: class {
+public protocol MiniAppMessageDelegate: class {
```
corycaywood, line 8: Won't this be a breaking change for the Host App because they have to rename it? Is it possible to keep `MiniAppMessageProtocol` for now and deprecate it?

I guess you would also need to add a deprecated `MiniApp.create` method for accepting `MiniAppMessageProtocol`.

Climbatize: No this won't be breaking change because I created a `typeAlias` :

```swift
@available(*, deprecated, message: "protocol renamed to MiniAppMessageDelegate")
public typealias MiniAppMessageProtocol = MiniAppMessageDelegate
```

I tested it without changing anything in the sample host app, everything is retro compatible

corycaywood: Ah, okay nice! Sorry, I didn't see that.

corycaywood (approved):
Looks good 👍

MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift
```diff
@@ -174,6 +184,18 @@ internal class MiniAppScriptMessageHandler: NSObject, WKScriptMessageHandler {
...
+        return ""
+    }
+
+    func requestPermission(permissionType: MiniAppPermissionType, completionHandler: @escaping (Result<MASDKPermissionResponse, MASDKPermissionError>) -> Void) {
+    }
+
+    func requestCustomPermissions(permissions: [MASDKCustomPermissionModel], completionHandler: @escaping (Result<[MASDKCustomPermissionModel], MASDKCustomPermissionError>) -> Void) {
+    }
+}
+
```
ts-leojoseph-r, line 198: extension is not `public`, which means the host app will get an error if it failed to conform to any `MiniAppMessageDelegate` methods. Is it the real intention here ?

MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift
```diff
@@ -127,23 +130,30 @@ internal class MiniAppScriptMessageHandler: NSObject, WKScriptMessageHandler {
...
-                    self.executeJavaScriptCallback(responseStatus: .onSuccess, messageId: callbackId, response: "SUCCESS")
-                case .failure(let error):
-                    if !error.localizedDescription.isEmpty {
-                        self.executeJavaScriptCallback(responseStatus: .onError, messageId: callbackId, response: error.localizedDescription)
-                        return
-                    }
-                    self.executeJavaScriptCallback(responseStatus: .onError, messageId: callbackId, response: getMiniAppErrorMessage(MiniAppErrorType.unknownError))
-                }
+            let info = MiniAppShareContent(messageContent: requestParamValue.content)
+            let hamDelegate = self.hostAppMessageDelegate ?? self
```
ts-leojoseph-r, line 134: I see the `hostAppMessageDelegate` value is not optional, why do we still trying to `self` ?

ts-leojoseph-r (commented):
Looks good to me :+1: Just few comments that need to clarified

MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift
```diff
@@ -174,6 +184,18 @@ internal class MiniAppScriptMessageHandler: NSObject, WKScriptMessageHandler {
     }
 }
 
+extension MiniAppScriptMessageHandler: MiniAppMessageDelegate {
+    func getUniqueId() -> String {
+        return ""
```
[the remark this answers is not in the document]
Climbatize: ok, no problem, that just feels weird for an ID :)

MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift
Climbatize: that's right, this is just for internal use. I implemented this because I wanted a fallback for this : `weak var hostAppMessageDelegate: MiniAppMessageDelegate?`
The weird think is that this protocol is optional, but you're right in your next comment, it also is a mandatory parameter, so maybe we can assume it will always be there and remove the optional

MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift
Climbatize: see my previous comment

MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift
Climbatize: I made it simpler, just assuming that there should not be an internal use of `MiniAppMessageDelegate`

ts-leojoseph-r (approved):
Looks good to me :+1:

corycaywood (approved):

PR #154, opened 2020-10-15
Feature/mini 287

Climbatize:
# Description
Updated sample App UI to filter MiniApps by name or by ID

## Links
[mini 287](https://jira.rakuten-it.com/jira/browse/MINI-287)

# Checklist


Example/Controllers/ViewController.swift
```diff
@@ -122,8 +125,34 @@ extension ViewController: UITableViewDelegate, UITableViewDataSource {
...
-        self.decodeResponse?.removeAll()
-        self.decodeResponse = miniAppList
+        self.unfilteredResults?.removeAll()
+        self.unfilteredResults = miniAppList
+        self.tableView.reloadData()
+    }
+}
+
+extension ViewController: UISearchBarDelegate {
+    func searchBarSearchButtonClicked(_ searchBar: UISearchBar) {
```
khairul-alam-licon, line 135: I am not sure, if only `searchButtonClicked` naming is good. Feel free about my suggestion :D

Climbatize: this is Apple UISearchBarDelegate protocol method, we should send them a ticket :)

khairul-alam-licon: Sorry, I [got it.](https://developer.apple.com/documentation/uikit/uisearchbardelegate/1624294-searchbarsearchbuttonclicked?language=objc) Please ignore my comment.

ts-leojoseph-r (commented):
Looks like you clubbed both functionalities which is really good, but I am not sure how QA is going to respond. Looks good. Few cosmetic comments may be.

1. Can you change the Title and placeholder text to white or some other color??
2. The table view and search bar is reset when you do pull to refresh, which shouln't be happening. I guess Android works the same way and fixed it recently.
Screen Shot 2020-10-16 at 12 25 41

Climbatize:
> Looks like you clubbed both functionalities which is really good, but I am not sure how QA is going to respond. Looks good. Few cosmetic comments may be.
> 
> 1. Can you change the Title and placeholder text to white or some other color??
> 2. The table view and search bar is reset when you do pull to refresh, which shouldn't be happening. I guess Android works the same way and they fixed it recently.
> 
> Screen Shot 2020-10-16 at 12 25 41

1. I can't do it, I mean it's not allowed by Apple. I changed to a more default color
2. OK no problem

khairul-alam-licon:
> The table view and search bar is reset when you do pull to refresh, which shouldn't be happening. I guess Android works the same way and they fixed it recently.

Yeah, in Android side "pull to refresh" option is disabled while searching because search-feature belongs to only downloaded MiniApps. I am not sure what is the best principle for iOS.

ts-leojoseph-r (approved):
Looks good to me :+1:

khairul-alam-licon (approved):
👍

PR #156, opened 2020-10-20
Fix for the memory leak

ts-leojoseph-r:
# Description
Fixing this memory leak will make the MiniApp object released when it is dismissed. This will fix the video-playing-in-background fix.

## Links
[MINI-1936](https://jira.rakuten-it.com/jira/browse/MINI-1936)

# Checklist

### DIFF
--- a/MiniApp/Classes/Display/MiniAppExternalWebViewController.swift
+++ b/MiniApp/Classes/Display/MiniAppExternalWebViewController.swift
@@ -14,7 +14,7 @@ class MiniAppExternalWebViewController: UIViewController {
     public class func presentModally(url: URL, externalLinkResponseHandler: MiniAppNavigationResponseHandler?) {
         let window: UIWindow?
         if #available(iOS 13, *) {
-            window = UIApplication.shared.windows.filter {$0.isKeyWindow}.first
+            window = UIApplication.shared.windows.filter { $0.isKeyWindow }.first
         } else {
             window = UIApplication.shared.keyWindow
         }
@@ -36,6 +36,7 @@ class MiniAppExternalWebViewController: UIViewController {
         config.preferences.javaScriptEnabled = true
         config.allowsInlineMediaPlayback = true
         config.allowsPictureInPictureMediaPlayback = true
+        config.mediaTypesRequiringUserActionForPlayback = []
 
         self.webView = WKWebView(frame: self.view.frame, configuration: config)
         self.webView.allowsBackForwardNavigationGestures = true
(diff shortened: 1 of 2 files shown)


corycaywood (approved):
👍

MiniApp/Classes/Display/MiniAppExternalWebViewController.swift
```diff
@@ -53,6 +54,11 @@ class MiniAppExternalWebViewController: UIViewController {
             self.webView.load(URLRequest(url: url))
         }
     }
+
+    override func viewDidDisappear(_ animated: Bool) {
+        super.viewDidDisappear(true)
```
Climbatize, line 59: you should use the animated parameter to call the super

khairul-alam-licon (approved):
👍

MiniApp/Classes/Display/MiniAppExternalWebViewController.swift
```diff
@@ -53,6 +54,11 @@ class MiniAppExternalWebViewController: UIViewController {
...
+        self.miniAppExternalUrlLoader = nil
```
Climbatize, line 60: this is working and it's cool but can't be a permanent fix, because if you loose the reference to `miniAppExternalUrlLoader`, viewdiddisappear does not guarantee your controller needs to be destroyed, the controller may be brought back to the user into a navigation for example, and then the MiniApp will never have the feedback.

Climbatize (commented):
I'm glad we found the same root of the problem. I have just some concerns about the way you fix it :)

MiniApp/Classes/Display/MiniAppExternalWebViewController.swift
ts-leojoseph-r: Done

Climbatize (changes requested):
@ts-leojoseph-r , @corycaywood please take a look at this fix: #157 and tell me what you think

corycaywood (commented):
I think the solution from @Climbatize might address the issue more fully, so let's look at that one.

MiniApp/Classes/Display/MiniAppExternalWebViewController.swift
ts-leojoseph-r: That's right. This wasn't a permanent fix. We need to realign the way External Webview is loaded. But as for this fix loosing `miniAppExternalUrlLoader` reference at `viewDidDisappear` will make the `MiniAppExternalWebViewController` release immediately. Not sure why `miniAppExternalUrlLoader` holding a strong reference, which I need to investigate further

Climbatize (approved):
much better, nice job

PR #160, opened 2020-10-22
Orientation lock support for Mini app

ts-leojoseph-r:
# Description
* Enable Orientation lock from the Mini app. Mini app can lock the orientation using the JS interfaces and the list of Orientation locks supported is,
    * Portrait
    * Landscape
* Updated sample app plist to support all orientations
* Updated sample app to enable Orientation Lock

## Links
[MINI-1751](https://jira.rakuten-it.com/jira/browse/MINI-1751)

# Checklist

### DIFF
--- a/Example/Controllers/DisplayController.swift
+++ b/Example/Controllers/DisplayController.swift
@@ -1,11 +1,14 @@
 import UIKit
 import MiniApp
+import AVKit
 
 class DisplayController: UIViewController {
 
     @IBOutlet var backButton: UIBarButtonItem!
     @IBOutlet var forwardButton: UIBarButtonItem!
     weak var navBarDelegate: MiniAppNavigationBarDelegate?
+    weak var miniAppDisplayDelegate: MiniAppDisplayProtocol?
+    static var miniAppSupportedOrientation: UIInterfaceOrientationMask = []
 
     override func viewDidAppear(_ animated: Bool) {
         guard let controller = self.navigationController as? DisplayNavigationController, let info = controller.miniAppInfo, let miniAppDisplay = controller.miniAppDisplay else {
@@ -13,11 +16,12 @@ class DisplayController: UIViewController {
         }
 
         self.title = info.displayName
-
+        self.miniAppDisplayDelegate = miniAppDisplay
         let view = miniAppDisplay.getMiniAppView()
         view.frame = self.view.bounds
         self.navBarDelegate = miniAppDisplay as? MiniAppNavigationBarDelegate
         self.view.addSubview(view)
+        self.navigationController?.delegate = self
     }
 
     @IBAction func done(_ sender: UIBarButtonItem) {
@@ -35,5 +39,22 @@ class DisplayController: UIViewController {
         default:
             break
         }
+    }
+
+    override var supportedInterfaceOrientations: UIInterfaceOrientationMask {
+        DisplayController.miniAppSupportedOrientation = self.miniAppDisplayDelegate?.getSupportedOrientation() ?? .all
+        return DisplayController.miniAppSupportedOrientation
+    }
+}
+
+extension DisplayController: UINavigationControllerDelegate {
+    public func navigationControllerSupportedInterfaceOrientations(_ navigationController: UINavigationController) -> UIInterfaceOrientationMask {
+        return navigationController.topViewController?.supportedInterfaceOrientations ?? .all
+    }
+}
+
+extension AVPlayerViewController {
+    override open var supportedInterfaceOrientations: UIInterfaceOrientationMask {
+        return DisplayController.miniAppSupportedOrientation
     }
 }
--- a/MiniApp/Classes/JavascriptBridge/MiniAppJavaScriptEnumerations.swift
+++ b/MiniApp/Classes/JavascriptBridge/MiniAppJavaScriptEnumerations.swift
@@ -6,6 +6,7 @@ enum MiniAppJSActionCommand: String {
     case shareInfo
     case getUserName
     case getProfilePhoto
+    case setScreenOrientation
 }
 
 enum JavaScriptExecResult: String {
@@ -52,6 +53,23 @@ public enum MiniAppCustomPermissionGrantedStatus: String, Codable {
         return true
         default:
         return false
+        }
+    }
+}
+
+enum MiniAppInterfaceOrientation: String, Codable {
+    case lockPortrait = "rakuten.miniapp.screen.LOCK_PORTRAIT"
+    case lockLandscape = "rakuten.miniapp.screen.LOCK_LANDSCAPE"
+    case lockRelease = "rakuten.miniapp.screen.LOCK_RELEASE"
+
+    public var orientation: UIInterfaceOrientationMask {
+        switch self {
+        case .lockPortrait:
+            return .portrait
+        case .lockLandscape:
+            return .landscape
+        case .lockRelease:
+            return .all
         }
     }
 }
--- a/MiniApp/Classes/JavascriptBridge/MiniAppJavaScriptEnumerations.swift
+++ b/MiniApp/Classes/JavascriptBridge/MiniAppJavaScriptEnumerations.swift
@@ -69,7 +69,7 @@ enum MiniAppInterfaceOrientation: String, Codable {
         case .lockLandscape:
             return .landscape
         case .lockRelease:
-            return .all
+            return []
         }
     }
 }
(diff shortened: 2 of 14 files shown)


MiniApp/Classes/JavascriptBridge/MiniAppJavaScriptEnumerations.swift
```diff
@@ -55,3 +56,20 @@ public enum MiniAppCustomPermissionGrantedStatus: String, Codable {
...
+    case lockRelease = "rakuten.miniapp.screen.LOCK_RELEASE"
+
+    public var orientation: UIInterfaceOrientationMask {
+        switch self {
+        case .lockPortrait:
+            return .portrait
+        case .lockLandscape:
+            return .landscape
+        case .lockRelease:
+            return .all
```
corycaywood, line 72: Could this override the Host App's default orientation setting? For example, this scenario: 
1. The App's view controller is locked to portrait mode by default
2. Then the mini app locks to landscape mode
3. Then the mini app releases the lock
4. The SDK sets the view controller to `.all`

Would this cause the App's view controller to no longer be locked to portrait and enable auto-rotate? Is there a way to save the App's original orientation and set it when the lock is released?

Example/Controllers/DisplayController.swift
```diff
@@ -36,4 +40,21 @@ class DisplayController: UIViewController {
...
+    }
+}
+
+extension DisplayController: UINavigationControllerDelegate {
+    public func navigationControllerSupportedInterfaceOrientations(_ navigationController: UINavigationController) -> UIInterfaceOrientationMask {
+        return navigationController.topViewController?.supportedInterfaceOrientations ?? .all
+    }
+}
+
+extension AVPlayerViewController {
```
corycaywood, line 56: This seems a little complicated that we have to ask the Host App to override the supported orientation interfaces for their view controller, navigation controller, and the avplayer. But I guess there isn't really any other way of doing this 😣

MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift
```diff
@@ -181,6 +184,23 @@ internal class MiniAppScriptMessageHandler: NSObject, WKScriptMessageHandler {
         let customPermission = self.miniAppKeyStore.getCustomPermissions(forMiniApp: self.miniAppId).filter { $0.permissionName == customPermissionType }
         return customPermission[0].isPermissionGranted.boolValue
     }
+
+    func setScreenOrientation(requestParam: RequestParameters?, callbackId: String) {
```
corycaywood, line 188: On Android, we have added a setting for `MiniAppMessageBridge.allowScreenOrientation`. So if the Host App doesn't enable screen orientation changes, then the SDK won't do anything when the mini app calls `setScreenOrientation`.

Does this make sense to add on iOS? If the Host App doesn't override `UIViewController. supportedInterfaceOrientations` will the SDK be able to change anything about the orientation?

corycaywood (commented):
It's looking good 👍 See my comments below

Example/Controllers/DisplayController.swift
ts-leojoseph-r: Yes, that's the sad part here. 

1. `override var supportedInterfaceOrientations: UIInterfaceOrientationMask`
 This is to override the supported orientation for the view controller i.e the display controller which the host app would use to display the mini-app. If the mini-app wants to lock the orientation for the viewcontroller(not just the video), they can do it only if the host app override this.
2. `public func navigationControllerSupportedInterfaceOrientations`
 DisplayController is pushed in the navigation controller and without this override, Viewcontroller is not locking the orientation. This is little strange for me too, when I tested in a small sample app, orientation lock is working fine without this override. I will verify it one more time.
3. `AVPlayerViewController - > override open var supportedInterfaceOrientations`
This is to override the orientation of the default media player. The host app can still skip this one, this is not mandatory. If the host app didn't override this, then the video wouldn't rotate automatically to landscape mode when the lock is set. User has to manually rotate the screen to lock in the landscape mode.

Example/Controllers/DisplayController.swift
```diff
@@ -1,23 +1,27 @@
 import UIKit
 import MiniApp
+import AVKit
```
Climbatize, line 3: the use of AVKit is really annoying to me. I know it's really convenient and maybe the only solution, but it's an heavy framework to just manage orientation

Climbatize (approved):
If we are sure we can't find another way than using AVKit I approve this PR (after you check Cory's comments of course). It's a very smart way to bypass Apple restrictions on the subject by the way

Example/Controllers/DisplayController.swift
corycaywood: Thanks, this is probably okay. About `AVPlayerViewController`, this is specifically what the orientation feature is needed for - to automatically set videos to landscape mode. 

Seems this will be confusing to the Host App, but it should be okay if we clearly document in the user guide how to handle orientation and what each override is for.

MiniApp/Classes/JavascriptBridge/MiniAppJavaScriptEnumerations.swift
ts-leojoseph-r: This won't override the host app default orientation, returning `.all` would reset the host app to its orientation once the lock is released. However, it is not good to return `.all` without knowing the host app current orientation. I have updated the code now to return the `.landscape` or `.portrait` based on the device rotation. Thank you 😄

Example/Controllers/DisplayController.swift
ts-leojoseph-r: Sure, I will add it in the documentation. I just noticed now that if the host app didn't override `AVPlayerViewController - > override open var supportedInterfaceOrientations` then the orientation lock is not working. This is because the Video player opened from the mini-app. I guess the host app need to implement this for sure, I will update all these things in the documentation once the PR is merged.

MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift
ts-leojoseph-r: I don't think we need this in iOS. We have implemented it differently in iOS due to the delegation pattern for orientation.

`func setScreenOrientation(requestParam: RequestParameters?, callbackId: String)` will only lock/set the orientation for the mini-app/video and the Host app must use `self.miniAppDisplayDelegate?.getSupportedOrientation()` in DisplayController or any viewcontroller which they are displaying Mini-app view.
And `self.miniAppDisplayDelegate?.getSupportedOrientation()` i.e MiniAppDisplayProtocol will retrieve the orientation supported for that mini-app.

Example/Controllers/DisplayController.swift
ts-leojoseph-r: Sadly yes, I don't like it either.
And if we don't use this, then the orientation lock is not working for the video player. Funny thing is, even if the app supports only portrait mode, the video player can rotate to all orientations(without this extension). So having this code in place helps us to lock the orientation.

corycaywood (approved):
Looks good 👍

MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift
```diff
@@ -181,6 +184,23 @@ internal class MiniAppScriptMessageHandler: NSObject, WKScriptMessageHandler {
...
+            executeJavaScriptCallback(responseStatus: .onError, messageId: callbackId, response: getMiniAppErrorMessage(MiniAppJavaScriptError.unexpectedMessageFormat))
+            return
+        }
+        if !requestParamValue.isEmpty {
+            guard let info = MiniAppInterfaceOrientation(rawValue: requestParamValue) else {
+                self.executeJavaScriptCallback(responseStatus: .onError, messageId: callbackId, response: getMiniAppErrorMessage(MiniAppJavaScriptError.unexpectedMessageFormat))
+                return
+            }
+            delegate?.didOrientationChanged(orientation: info.orientation)
+            UIViewController.attemptRotationToDeviceOrientation()
```
corycaywood, line 199: Sorry, I missed something here. We need to send a `success` callback to the mini app. The Android SDK sends success after then orientation was successfully set.

It looks like there isn't a way to detect if `attemptRotationToDeviceOrientation` was successful? So maybe it's okay just to send success immediately after calling `attemptRotationToDeviceOrientation`?

ts-leojoseph-r: Oh, sorry. I thought Android isn't sending any success response, I looked at few android comments. No probs, I will make changes in another PR 👍

PR #161, opened 2020-10-28
Feat/access token

ts-leojoseph-r:
# Description
* Retrieve access token and expiration date from the Host app.

## Links
[MINI-1864](https://jira.rakuten-it.com/jira/browse/MINI-1864)

# Checklist


MiniApp/Classes/Protocols/MiniAppAuthenticationDelegate.swift
```diff
@@ -0,0 +1,24 @@
+/**
+ Public Protocol that will be used by the Mini App to get Authentication
+ info, such as Access token
+*/
+public protocol MiniAppAuthenticationDelegate: class {
+    /// Interface that is used to retrieve the Token Info
+    func getAccessToken(completionHandler: @escaping (Result<MATokenInfo, MASDKCustomPermissionError>) -> Void)
```
corycaywood, line 7: We need to also pass the mini app ID to this method. This is so the Host App can verify that this mini app should be able to use the access token. See the android implementation: 
https://github.com/rakutentech/android-miniapp/blob/master/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/userinfo/UserInfoBridgeDispatcher.kt#L30

MiniApp/Classes/Protocols/MiniAppAuthenticationDelegate.swift
```diff
@@ -0,0 +1,24 @@
+/**
+ Public Protocol that will be used by the Mini App to get Authentication
+ info, such as Access token
+*/
+public protocol MiniAppAuthenticationDelegate: class {
```
corycaywood, line 5: For Android, we added this to `UserInfoBridgeDispatcher`. So maybe the `MiniAppUserInfoDelegate` is a better place for this? This is related to user info, so I think it makes more sense.

MiniApp/Classes/JavascriptBridge/MiniAppJavaScriptEnumerations.swift
```diff
@@ -38,6 +40,8 @@ public enum MiniAppCustomPermissionType: String, Codable, CaseIterable {
         return "Profile Photo"
         case .contactsList:
         return "Contact List"
+        case .accessToken:
```
corycaywood, line 43: Sorry, I forgot to update you about this 😣  We don't want to add a permission for the access token yet. Instead, we will pass the Mini App ID to the Host App and let them verify that the mini app should be able to use the access token.

This is because the access token won't be used by all Host Apps, so we don't want the permission to show up in the permission management screen. In the future, we will change the permissions management functionality so it returns only the permissions which the mini app is actually using (as set in their JSON manifest). Then we can add a permission for access token and anything else that won't be used by all Host Apps.

corycaywood (commented):
Looking good, but I forgot to update the requirements after we made some changes to the Android side. Sorry about that 😣 

Also, we should let the access token and expiration date be settable via the Demo App's settings screen. But probably you can do that in another PR if  you want.

MiniApp/Classes/Protocols/MiniAppAuthenticationDelegate.swift
ts-leojoseph-r: Done 👍

MiniApp/Classes/JavascriptBridge/MiniAppJavaScriptEnumerations.swift
ts-leojoseph-r: Understood 👍

MiniApp/Classes/Protocols/MiniAppAuthenticationDelegate.swift
ts-leojoseph-r: I thought having it as a separate delegate would be better. As of now, we retrieve only token and expiryDate and I thought we will have more into R SDK in the future. No worries, updated 👍

corycaywood (approved):
LGTM 👍  As I mentioned, just need to make the access token configurable in the demo app settings, but that can be done in another PR.

khairul-alam-licon (approved):
👍

corycaywood:
@leorjoseph The access token features have been merged for the JS bridge in `js-miniapp`, so you should also update the JS submodule before merging this.

MiniApp/Classes/Protocols/MiniAppUserInfoDelegate.swift
```diff
@@ -18,4 +21,18 @@ public extension MiniAppUserInfoDelegate {
...
     }
+
+    func getAccessToken(miniAppId: String, completionHandler: @escaping (Result<MATokenInfo, MASDKCustomPermissionError>) -> Void) {
+        completionHandler(.failure(.failedToConformToProtocol))
+    }
+}
+
+public class MATokenInfo: Codable {
+    let token: String
+    let validUntil: Date
```
corycaywood, line 32: Is `Date` the correct format for this? The JS bridge is expecting a `number`. I think it's okay to have the Host App provide a `Date`, but can we convert it to the correct number format before sending to the bridge?

MiniApp/Classes/Protocols/MiniAppUserInfoDelegate.swift
```diff
@@ -18,4 +21,18 @@ public extension MiniAppUserInfoDelegate {
...
+
+    public init(accessToken: String, expirationDate: Date) {
+        self.token = accessToken
+        self.validUntil = expirationDate
```
corycaywood, line 36: Regarding my above comment, basically here could you set `self.validUntil = expirationDate.timeIntervalSince1970`?

Climbatize (approved):

MiniApp/Classes/Protocols/MiniAppUserInfoDelegate.swift
ts-leojoseph-r: Updated now

MiniApp/Classes/Protocols/MiniAppUserInfoDelegate.swift
ts-leojoseph-r: expirationDate.timeIntervalSince1970 returns a Double. I have updated the code to return Number `Int` now.

corycaywood (approved):
LGTM 👍

PR #162, opened 2020-10-28
Update how host app manages Orientation lock 

ts-leojoseph-r:
# Description
* Updated Host app to manage Orientation lock via AppDelegate
* Updated code to send success response once the orientation lock is set
* Few code refactoring

# Checklist

### DIFF
--- a/USERGUIDE.md
+++ b/USERGUIDE.md
@@ -365,7 +365,7 @@ func application(_ application: UIApplication, supportedInterfaceOrientationsFor
     }
 ```
 
-### Allow only full screen videos to change orientation
+### Allow full screen videos to change orientation
 
 If you do not wish to add code to `AppDelegate` as in the above example, you can still allow videos inside the mini app to use landscape mode even when your App is locked to portrait mode and vice versa.
 
--- a/USERGUIDE.md
+++ b/USERGUIDE.md
@@ -367,7 +367,7 @@ func application(_ application: UIApplication, supportedInterfaceOrientationsFor
 
 ### Allow full screen videos to change orientation
 
-If you do not wish to add code to `AppDelegate` as in the above example, you can still allow videos inside the mini app to use landscape mode even when your App is locked to portrait mode and vice versa.
+You can add the following if you want to enable videos to change orientation. Note that if you do not wish to add code to `AppDelegate` as in the above example, you can still allow videos inside the mini app to use landscape mode even when your App is locked to portrait mode and vice versa.
 
 ```swift
 import AVKit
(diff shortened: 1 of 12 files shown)


corycaywood (commented):
Nice, this looks good 👍 

You also need to update the user guide with this information. But I think we should split it into two sections so they can choose to not add anything to `AppDelegate` if they don't want to. For Example:

```
## Orientation

You can choose to give orientation lock control to mini apps. However, this requires you to add some code to your `AppDelegate` which could have an affect on your entire App. If you do not wish to do this, please see the section "Allow only full screen videos to change orientation".

### Allow mini apps to lock the view to any orientation

......

### Allow only full screen videos to change orientation

If you do not wish to add code to `AppDelegate` as in the above example, you can still allow videos inside the mini app to use landscape mode even when your App is locked to portrait mode and vice versa.

import AVKit

extension AVPlayerViewController {
    open override var supportedInterfaceOrientations: UIInterfaceOrientationMask {
        return .all
    }
}
```

What do you think?

corycaywood (approved):
Looks good! 👍

USERGUIDE.md
```diff
@@ -345,50 +345,43 @@ extension ViewController: MiniAppMessageDelegate {
...
         }
     }
-    ```
-    NOTE: You should also set the Navigation Controller delegate to self. ```self.navigationController?.delegate = self```
-
-3. Overriding `supportedInterfaceOrientations` for AVPlayerViewController
+```
 
-    Orientation lock will not work for videos that is played inside Mini-app. To enable Orientation lock for the videos, please add the following code,
+### Allow only full screen videos to change orientation
```
corycaywood, line 368: I know that this is what I wrote, but isn't this always required if they want landscape video? So maybe we should remove `only`?

```suggestion
### Allow full screen videos to change orientation
```

USERGUIDE.md
```diff
@@ -345,50 +345,43 @@ extension ViewController: MiniAppMessageDelegate {
...
 
-    `import AVKit`
+If you do not wish to add code to `AppDelegate` as in the above example, you can still allow videos inside the mini app to use landscape mode even when your App is locked to portrait mode and vice versa.
```
corycaywood, line 370: Same here, maybe this text needs to be a little different to clarify that it's always required for video?

```suggestion
You can add the following if you want to enable videos to change orientation. Note that if you do not wish to add code to `AppDelegate` as in the above example, you can still allow videos inside the mini app to use landscape mode even when your App is locked to portrait mode and vice versa.
```

Climbatize (approved):
LGTM

PR #163, opened 2020-11-09
Feat/custom permissions ui extensions

ts-leojoseph-r:
# Description
* Updated SDK to have a default implementation of Custom Permissions Request flow.

## Links

- [MINI-1679](https://jira.rakuten-it.com/jira/browse/MINI-1679)
- [MINI-1913](https://jira.rakuten-it.com/jira/browse/MINI-1913)

# Checklist


MiniApp/Classes/Protocols/MiniAppMessageDelegate.swift
```diff
@@ -14,11 +14,45 @@ public protocol MiniAppMessageDelegate: MiniAppUserInfoDelegate, MiniAppShareCon
...
+    /// Optional Interface that can be implemented in the host app to handle the Custom Permissions.
+    /// SDK will open its default UI for prompting Custom Permissions request if this protocol interface is not overridden
+    func requestCustomPermissions(permissions: [MASDKCustomPermissionModel],
+                                  miniAppTitle: String,
+                                  completionHandler: @escaping (Result<[MASDKCustomPermissionModel], MASDKCustomPermissionError>) -> Void)
+
     /// Interface that should be implemented in the host app to handle the Custom Permissions.
     /// Host app is responsible to display the alert/dialog with the [MiniAppCustomPermissionType] permissions to the user and the result should be returned back to the SDK
+    @available(*, deprecated,
+    message:"Since version 2.3.0, you can requestCustomPermissions using the below method",
```
Climbatize, line 26: "the below method" makes no sense when autocompleting in Xcode. the renamed section gives enough informations about the method to switch to

Climbatize (approved):
LGTM

MiniApp/Classes/Protocols/MiniAppMessageDelegate.swift
ts-leojoseph-r: I happen to remove this method now. I was thinking about whether we need to provide a deprecated note for this protocol? If yes, then we need to call both old and new delegates from https://github.com/rakutentech/ios-miniapp/pull/163/files#diff-eaffa51bae0dd77ddb253a0469ff68a7914e6f8ff719a16a2638b706f0d64059L65. What do you think ?

ts-leojoseph-r: Updated

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 **SDK**
 
 - **Feature:** Added support for Orientation lock, that enables the mini app to lock `portrait` or `landscape` orientation for the mini-app. [Please check here](USERGUIDE.md#orientation-lock)
+- **Feature:** Added default implementation in SDK for `requestCustomPermissions(permissions:miniAppTitle:completionHandler:)` [Please check here](USERGUIDE.md#request-custom-permission)
+- **Deprecated:** `requestCustomPermissions(permissions:completionHandler:)` in `MiniAppCallbackProtocol` protocol is deprecated
```
khairul-alam-licon, line 9: One qus, do we need to mark the interface as deprecated? In my understanding, HostApp still can use it when they want to implement their own UI.

corycaywood: I think that this is deprecated because it's been replaced by a new method - so it is still possible for the host app to override it.

@leorjoseph Maybe add something like this here: `You should use requestCustomPermissions(permissions: miniAppTitle:completionHandler:) instead`

corycaywood (approved):
Looks good 👍  Just one note on the changelog.

khairul-alam-licon (approved):
Looks good 👍

CHANGELOG.md
ts-leojoseph-r: @corycaywood It is already added to the developer documentation. Sure, I will add it here as well 👍

PR #165, opened 2020-11-12
feat: Return custom Error type to clients

corycaywood:
# Description
The basic idea is that we wanted to allow the Host App to handle the error cases for 'no published version' and 'mini app not found'. But I found that the current implementation for returning Errors doesn't really allow them to easily do that - they would have to convert the error to `NSError` and check the error domain and error code.

I wanted to make it so the Host App can easily handle errors in the Swift style by using switch/case, so I've changed the `MiniApp` functions to return a new error type `MASDKError`. The `NSError` will be converted to `MASDKError` before being returned to the Host App.

One problem with this is that it would be a breaking change for Host Apps if they've explicitly defined their completion handler to use an `Error`. So the method I found to get around this is to add deprecated methods which use `where T: Error`. This will mean that the functions using `MASDKError` will be used by default if the Host App has not defined an explicit type, but the deprecated functions can still be used if the Host App is explicitly using `Error`.

## Links
MINI-1972

# Checklist

### DIFF
--- a/Example/Controllers/Extensions/ViewController+MiniApp.swift
+++ b/Example/Controllers/Extensions/ViewController+MiniApp.swift
@@ -34,9 +34,18 @@ extension ViewController: MiniAppNavigationDelegate {
                     self.currentMiniAppInfo = responseData
                     self.fetchMiniApp(for: responseData)
                 case .failure(let error):
+                    var message: String
+                    switch error {
+                    case .noPublishedVersion:
+                        message = NSLocalizedString("error_no_published_version", comment: "")
+                    case .miniAppNotFound:
+                        message = NSLocalizedString("error_miniapp_id_not_found", comment: "")
+                    default:
+                        message = NSLocalizedString("error_single_message", comment: "")
+                    }
                     print(error.localizedDescription)
                     self.dismissProgressIndicator {
-                        self.fetchMiniAppUsingId(title: NSLocalizedString("error_title", comment: ""), message: NSLocalizedString("error_single_message", comment: ""))
+                        self.fetchMiniAppUsingId(title: NSLocalizedString("error_title", comment: ""), message: message)
                     }
                 }
             }
--- a/MiniApp/Classes/MASDKError.swift
+++ b/MiniApp/Classes/MASDKError.swift
@@ -8,26 +8,26 @@ public enum MASDKError: Error {
     ///     - code: HTTP status code from the server
     ///     - message: error message returned by the server
     case serverError(code: Int, message: String)
-    
+
     /// The URL was invalid the SDK tried to connect to the server.
     /// Probably you provided an invalid URL value to the Base URL setting.
     case invalidURLError
-    
+
     /// The provided mini app ID was invalid. For example, the value cannot be an empty string.
     case invalidAppId
-    
+
     /// The server provided an invalid response body.
     case invalidResponseData
-    
+
     /// The mini app failed to download.
     case downloadingFailed
-    
+
     /// There are no published versions for the provided mini app ID.
     case noPublishedVersion
-    
+
     /// The provided mini app ID was not found on the server.
     case miniAppNotFound
-    
+
     /// An unexpected error occured.
     ///
     /// - Parameters:
--- a/MiniApp/Classes/MASDKError.swift
+++ b/MiniApp/Classes/MASDKError.swift
@@ -1,0 +1,89 @@
+import Foundation
+
+/// Mini App SDK Error which is passed from the main MiniApp APIs.
+public enum MASDKError: Error {
+    /// The server returned an error.
+    ///
+    /// - Parameters:
+    ///     - code: HTTP status code from the server
+    ///     - message: error message returned by the server
+    case serverError(code: Int, message: String)
+    
+    /// The URL was invalid the SDK tried to connect to the server.
+    /// Probably you provided an invalid URL value to the Base URL setting.
+    case invalidURLError
+    
+    /// The provided mini app ID was invalid. For example, the value cannot be an empty string.
+    case invalidAppId
+    
+    /// The server provided an invalid response body.
+    case invalidResponseData
+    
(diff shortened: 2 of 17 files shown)


corycaywood:
@leorjoseph @Climbatize Please see the PR description above. This is just a draft PR for now, and I can update the unit tests, etc later.

I'm still learning about error handling in Swift and iOS, so I'm not sure if this is the best approach. So please let me know your comments and suggestions.

Example/Controllers/Extensions/ViewController+MiniApp.swift
```diff
@@ -34,9 +34,18 @@ extension ViewController: MiniAppNavigationDelegate {
                     self.currentMiniAppInfo = responseData
                     self.fetchMiniApp(for: responseData)
                 case .failure(let error):
+                    var message: String
+                    switch error {
+                    case .noPublishedVersion:
+                        message = "No published versions for the provided Mini App ID."
+                    case .miniAppNotFound:
+                        message = "No Mini App found for the provided Mini App ID."
```
Climbatize, line 42: if we localized the regular message maybe we can also localize these two

Climbatize (approved):
LGTM

MiniApp/Classes/RealMiniApp.swift
```diff
@@ -146,6 +146,17 @@ internal class RealMiniApp {
     func getDownloadedListWithCustomPermissions() -> MASDKDownloadedListPermissionsPair {
         return self.miniAppStatus.getMiniAppsListWithCustomPermissionsInfo() ?? []
     }
+    
+    func createMASDKErrorCompletionHandler<T>(completionHandler: @escaping (Result<T, MASDKError>) -> Void) -> (Result<T, Error>) -> Void {
```
ts-leojoseph-r, line 150: Should we rename it to some other name? Because it handles both success and failure completion handlers.

MiniApp/Classes/MiniApp.swift
```diff
@@ -95,4 +94,50 @@ public class MiniApp: NSObject {
     public func create(appInfo: MiniAppInfo, completionHandler: @escaping (Result<MiniAppDisplayProtocol, Error>) -> Void, messageInterface: MiniAppMessageDelegate) {
         return realMiniApp.createMiniApp(appInfo: appInfo, completionHandler: completionHandler, messageInterface: messageInterface)
     }
+
+    @available(*, deprecated,
+    message:"Use MASDKError instead of Error in your completionHandler.",
+    renamed: "list(completionHandler:)")
+    public func list<T>(completionHandler: @escaping (Result<[MiniAppInfo], T>) -> Void) where T: Error {
```
ts-leojoseph-r, line 101: I guess this is added by mistake?

@availble syntax indicates that the method is deprecated but we weren't using 
`public func list<T>(completionHandler: @escaping (Result<[MiniAppInfo], T>) -> Void) where T: Error {` before.

The method which we currently support 
`public func list(completionHandler: @escaping (Result<[MiniAppInfo], Error>) -> Void) {`

MiniApp/Classes/MiniApp.swift
```diff
@@ -32,10 +32,11 @@ public class MiniApp: NSObject {
     ///     -   completionHandler: A block to be called when MiniAppInfo information is fetched. Completion blocks receives the following parameters
     ///         -   MiniAppInfo: MiniAppInfo information.
     ///         -   Error: Error details if fetching is failed.
-    public func info(miniAppId: String, completionHandler: @escaping (Result<MiniAppInfo, Error>) -> Void) {
+    public func info(miniAppId: String, completionHandler: @escaping (Result<MiniAppInfo, MASDKError>) -> Void) {
         if miniAppId.count == 0 {
-            return completionHandler(.failure(NSError.invalidAppId()))
+            return completionHandler(.failure(MASDKError.invalidAppId))
```
ts-leojoseph-r, line 37: You can return just `.invalidAppId`. We don't need to use `MASDKError.invalidAppId`

ts-leojoseph-r (changes requested):
Overall looks good. I would be happier to merge this, I knew the public interfaces and protocols were poorly managed at first. 
I had done the same for the `MiniAppMessageDelegate` protocols if you remember but we decided not to modify the existing public protocols and modified only the new protocols.

Added few comments, but other than that, it looks great 👍

MiniApp/Classes/MASDKError.swift
```diff
@@ -0,0 +1,63 @@
...
+        case .invalidResponseData:
+            return "Invalid response received from server."
+        case .downloadingFailed:
+            return "Failed to download the mini app."
+        case .noPublishedVersion:
+            return "Server returned no published versions for the provided Mini App ID."
+        case .miniAppNotFound:
+            return "Server could not find the provided Mini App ID."
+        case .unknownError(let domain, let code, let description):
+            return "Unknown error occured in \(domain) domain with error code \(code): \(description)"
```
ts-leojoseph-r, line 32: I guess these messages need to be localized.

Example/Controllers/Extensions/ViewController+MiniApp.swift
ts-leojoseph-r: I guess this is for the sample app, so I guess it's fine.

corycaywood: I localized these.

MiniApp/Classes/MiniApp.swift
corycaywood: Oh, cool. Thanks, I didn't know about that.

MiniApp/Classes/MiniApp.swift
corycaywood: This one is a little weird, but it was the only way I could find to support `MASDKError` as the new default but not break support of `Error` if the host app has explicitly defined that type.

ts-leojoseph-r (approved):
Looks good to me :+1:

PR #173, opened 2020-11-19
Updated sample app to use Project ID title in the settings

ts-leojoseph-r:
# Description
Updated in sample app settings to use `Project ID title`

# Checklist

### DIFF
--- a/Example/Info.plist
+++ b/Example/Info.plist
@@ -25,7 +25,7 @@
 	<key>NSLocationWhenInUseUsageDescription</key>
 	<string>Demo app helps you find more information near you</string>
 	<key>RASProjectId</key>
-	<string>$(RAS_PROJECT_ID)</string>
+	<string>$(RAS_APPLICATION_IDENTIFIER)</string>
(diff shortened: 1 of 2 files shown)


Example/Info.plist
```diff
@@ -25,7 +25,7 @@
 	<key>NSLocationWhenInUseUsageDescription</key>
 	<string>Demo app helps you find more information near you</string>
 	<key>RASProjectId</key>
-	<string>$(RAS_APPLICATION_IDENTIFIER)</string>
+	<string>$(RAS_PROJECT_ID)</string>
```
corycaywood, line 28: I think you can keep this as `RAS_APPLICATION_IDENTIFIER` for now. It also needs to be updated in the `configure-secrets.sh` script and then updated in all of the CI settings and on our local machines. So it will be kind of a pain to change 😣  I think we can handle this later.

ts-leojoseph-r: Ok 👍

corycaywood (approved):
Looks good 👍

Climbatize (approved):

PR #174, opened 2020-11-20
Quick fixes on Preview Feature in Sample app

Climbatize:
# Description

- Quick fixes on Preview Feature in Sample app
- Previewable apps were not loaded

# Checklist


ts-leojoseph-r (approved):
Looks good to me :+1:

corycaywood (approved):
Looks good 👍  I also updated the CI settings to add `RAS_PROJECT_IDENTIFIER`.

khairul-alam-licon (approved):
👍

MiniApp/Classes/MiniApp.swift
```diff
@@ -89,9 +90,6 @@ public class MiniApp: NSObject {
     ///                         to interact with View component of mini app.
     ///         -   Error: Error details if Mini App View creating is failed
     ///   - messageInterface: Protocol implemented by the user that helps to communicate between Mini App and native application
-    @available(*, deprecated,
-    message:"Since version 2.0, you can create a Mini app view using just the mini app id",
-    renamed: "create(appId:completionHandler:messageInterface:)")
     public func create(appInfo: MiniAppInfo, completionHandler: @escaping (Result<MiniAppDisplayProtocol, Error>) -> Void, messageInterface: MiniAppMessageDelegate) {
```
corycaywood, line 93: On Android, we updated the docs to mention that this method uses the version id from MiniAppInfo and also that it should only be used in preview mode. See here: https://github.com/rakutentech/android-miniapp/blob/master/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt#L54

Maybe we add something similar here?

MiniApp/Classes/RealMiniApp.swift
```diff
@@ -66,8 +66,8 @@ internal class RealMiniApp {
             } }
     }
 
-    func createMiniApp(appId: String, completionHandler: @escaping (Result<MiniAppDisplayProtocol, Error>) -> Void, messageInterface: MiniAppMessageDelegate? = nil) {
-        getMiniApp(miniAppId: appId) { (result) in
+    func createMiniApp(appId: String, version: String? = nil, completionHandler: @escaping (Result<MiniAppDisplayProtocol, Error>) -> Void, messageInterface: MiniAppMessageDelegate? = nil) {
```
corycaywood, line 69: I think it's okay for now, but an improvement for the future would be caching the files for multiple versions in preview mode. 

On Android, the downloader will check if it's in preview mode and only delete the cached version if it's not. So in published mode, it can only have one version downloaded at a time, but in preview mode it can have many versions downloaded. See here: https://github.com/rakutentech/android-miniapp/blob/f5f4c7d40751ba9416be88ec2018296a031d3735/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt#L115

MiniApp/Classes/MiniApp.swift
Climbatize: sure you're right

MiniApp/Classes/RealMiniApp.swift
Climbatize: I've done something similar in a separate PR #175

PR #177, opened 2020-11-25
Fix: Unit tests failing on slow CI machine

Climbatize:
# Description
We think that increasing timeout on some file manipulating test cases can solve intermittent issues experienced only during CI tests.

# Checklist


Tests/Unit/MiniAppDownloaderTests.swift
```diff
@@ -84,8 +84,8 @@ class MiniAppDownloaderTests: QuickSpec {
                      let oldMiniAppDirectory = FileManager.getMiniAppVersionDirectory(with: appId, and: "\(versionId).1")
                      var isDir: ObjCBool = true
 
-                     expect(FileManager.default.fileExists(atPath: miniAppDirectory.path, isDirectory: &isDir)).toEventually(equal(true), timeout: .seconds(10))
-                     expect(FileManager.default.fileExists(atPath: oldMiniAppDirectory.path, isDirectory: &isDir)).toEventually(equal(false), timeout: .seconds(10))
+                     expect(FileManager.default.fileExists(atPath: miniAppDirectory.path, isDirectory: &isDir)).toEventually(equal(true), timeout: .seconds(30))
```
corycaywood, line 87: 30s seems really long 😣 Is is possible to mock FileManager on iOS? Maybe in the future we can try that?

corycaywood (approved):
👍

Tests/Unit/MiniAppDownloaderTests.swift
ts-leojoseph-r: That's a max timeout for that testcase anyway, it will pass immediately if the expected value is returned.

ts-leojoseph-r (approved):
Looks good to me :+1:
Mini_App_view__when_RealMiniAppView_is_called_with_coder__will_return_nil testcase can be removed.

PR #181, opened 2020-11-27
Feat/location custom permissions

ts-leojoseph-r:
# Description
* Added Location permissions into Custom permissions
* If the user allow Device Location custom permissions only then the host app will prompt the device Location permission to the user.

## Links
MINI-1955

# Checklist

### DIFF
--- a/MiniApp/Classes/JavascriptBridge/MiniAppJavaScriptEnumerations.swift
+++ b/MiniApp/Classes/JavascriptBridge/MiniAppJavaScriptEnumerations.swift
@@ -30,6 +30,7 @@ public enum MiniAppCustomPermissionType: String, Codable, CaseIterable {
     case userName = "rakuten.miniapp.user.USER_NAME"
     case profilePhoto = "rakuten.miniapp.user.PROFILE_PHOTO"
     case contactsList = "rakuten.miniapp.user.CONTACT_LIST"
+    case deviceLocation = "rakuten.miniapp.device.LOCATION"
 
     public var title: String {
         switch self {
@@ -39,6 +40,8 @@ public enum MiniAppCustomPermissionType: String, Codable, CaseIterable {
             return "Profile Photo"
         case .contactsList:
             return "Contact List"
+        case .deviceLocation:
+            return "Device Location"
         }
     }
 }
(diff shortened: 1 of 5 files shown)


MiniApp/Classes/JavascriptBridge/MiniAppJavaScriptEnumerations.swift
```diff
@@ -39,6 +40,8 @@ public enum MiniAppCustomPermissionType: String, Codable, CaseIterable {
             return "Profile Photo"
         case .contactsList:
             return "Contact List"
+        case .deviceLocation:
+            return "Device Location"
```
khairul-alam-licon, line 44: 👍 I'll update this text in Android.

MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift
```diff
@@ -89,10 +89,14 @@ internal class MiniAppScriptMessageHandler: NSObject, WKScriptMessageHandler {
             return
         }
 
-        switch requestPermissionType {
-        case .location:
-            getPermissionResult(requestPermissionType: requestPermissionType, callbackId: callbackId)
+        if isUserAllowedPermission(customPermissionType: MiniAppCustomPermissionType.deviceLocation) {
```
corycaywood, line 92: I think we shouldn't change the behavior that we currently have here for the device permission. It makes more sense to me if we leave `requestPermission` to purely handle device permissions so we have a clear separation between custom and device permissions. For example, the host app is able to override the `requestPermission` functionality so it would be more complex if they have to worry about both permission types here, and we won't be able to predict from the JS SDK side how they are handling it.

I think it's better to handle requesting the custom permission/device permission together from the JS SDK side. See the PR from @khairul-alam-licon for the JS SDK: https://github.com/rakutentech/js-miniapp/pull/75/files

Do you think this makes sense?

ts-leojoseph-r: Gotcha 👍 , I didn't have a look into that ticket, sorry 👍

corycaywood (approved):
Looks good 👍

Climbatize (approved):
LGTM

PR #185, opened 2020-12-02
Refactor the documentation theme and contents

ts-leojoseph-r:
# Description
* Used a different theme (from Jazzy) for the documentation -> This actually solves lot of alignment issues and it looks much more cleaner than the previous one.
  * This will be the new Theme -> [Check here](https://reduxkit.github.io/ReduxKit/master/)
* Refactored documentation

# Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,7 @@ after_success:
   --module MiniApp
   --source-directory MiniApp
   --podspec MiniApp.podspec
+  --theme fullwidth
   --readme USERGUIDE.md
 # Deploy Simulator build to App Center when merged to master (this script doesn't work from the `deploy` section)
 # Deploys to STG group if merging to master or deploys to prod group if this is a release
--- a/DEV.md
+++ b/DEV.md
@@ -2,13 +2,25 @@
 
 This guide targets Developers of this SDK. It explains how to build the project and how the CI setup works. In order to setup the environment you need to install [Cocoapods](https://guides.cocoapods.org/using/getting-started.html) and [Fastlane](https://docs.fastlane.tools/getting-started/ios/setup/)
 
-## <a name="sample-app" /> How to Build the Sample App
+## Table of Contents
+---
 
-First, it is highly recommanded to setup some environment variables which are required by the project.
+* [How to Build the Sample App](#build-sample-app)
+* [How to Test with the Sample App](#test-sample-app)
+* [How to test your SDK integration with Cocoapods](#integrate-sdk-cocoapod)
+* [How to generate SDK documentation locally](#generate-sdk-doc)
+* [Continuous Integration and Deployment](#continuous-integration)
+
+<div id="build-sample-app"></div>
+
+## How to Build the Sample App
+---
+
+First, it is highly recommended to setup some environment variables which are required by the project.
 
 ```bash
 RMAAPIEndpoint=https://www.example.com
-RASProjectId=test-app-id
+RASProjectId=test-project-id
 RASProjectSubscriptionKey=test-subscription-key
 ```
 
@@ -18,11 +30,17 @@ Next, run `fastlane updatePods` from the root directory that will trigger a `pod
 
 *Note:* You need to define the environment variables before installing the pods because there is a post install script which sets up the project with your environment variables. If you don't want to use environment variables, you can edit the `MiniApp-Secrets.xcconfig` created after a `pod install` in the parent folder of the project,but be aware you will have to update this file after every `pod install`
 
+<div id="test-sample-app"></div>
+
 ## How to Test with the Sample App
+---
 
 We currently don't provide an API for public use, so you must provide your own API.
 
+<div id="integrate-sdk-cocoapod"></div>
+
 ## How to test your SDK integration with Cocoapods
+---
 
 If you need to test your SDK fork into your host app before making a pull request, you can use this line into your podfile:
 
@@ -30,7 +48,10 @@ If you need to test your SDK fork into your host app before making a pull reques
   pod 'MiniApp', git: 'https://github.com/<My fork account>/ios-miniapp', branch: 'master', submodules: true
 ```
 
+<div id="generate-sdk-doc"></div>
+
 ## How to generate SDK documentation locally
+---
 
 You may want to generate the SDK documentation locally so that you can ensure that the generated docs look correct. 
 We use [Jazzy](https://github.com/realm/jazzy) for this, so you can run the following commands:
@@ -46,7 +67,10 @@ bundle exec jazzy \
 
 The generated docs will be output to a folder named `docs` in the root of this repo.
 
+<div id="continuous-integration"></div>
+
 ## Continuous Integration and Deployment
+---
 
 Before any deployment, be sure the project will build and run unit tests by running `fastlane ci`.
 
--- a/DEV.md
+++ b/DEV.md
@@ -11,7 +11,7 @@ This guide targets Developers of this SDK. It explains how to build the project 
 * [How to generate SDK documentation locally](#generate-sdk-doc)
 * [Continuous Integration and Deployment](#continuous-integration)
 
-<div id="build-sample-app"></div>
+<a id="build-sample-app"></a>
 
 ## How to Build the Sample App
 ---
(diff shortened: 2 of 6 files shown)


Climbatize (approved):
Well done

corycaywood (commented):
The new theme looks really nice!

But I think there's still some work to be done on re-organizing the user guide. Specifically, I think we should make the structure more closely match Android ([see here](https://github.com/rakutentech/android-miniapp/blob/master/miniapp/USERGUIDE.md)). It doesn't have to exactly match, but I think it should be more similar.

Here are my thoughts on changes:
- Headers underneath "Getting Started" should be sub-headers - i.e. `## Requirements` -> `### Requirements`
- Maybe the "Usage" section can be rearranged to more closely match the Android guide. i.e. rearrange it to have sub-sections like this:
  - "Create a Mini App"
  - "Mini App Features" - contains info about communicating with mini app - `MiniAppMessageDelegate`, User Profile Details, Orientation Lock, etc.
  - "Fetching Mini App Info" - contains "Load the MiniApp list:" and "Getting MiniAppInfo"
  - "Custom Permissions"
  - "Advanced Features" - anything else: ""Add a web navigation interface to the MiniApp view"", "Overriding config at runtime", "Allow full screen video", etc.
- I think some sections should have more detailed information on what the functionality does. See Android's "Create and display a Mini App" and  "Mini App Features" section, for example (maybe some of this text can just be copied directly?)
- iOS already has some direct links to API docs - but I also like the idea of adding the relevant API Docs at the top of each section (see Android). Maybe that would work here too?

Sorry for the long list 😣  

@ts-leojoseph-r  @Climbatize  What do you think?

DEV.md
```diff
@@ -18,19 +30,28 @@ Next, run `fastlane updatePods` from the root directory that will trigger a `pod
 
 *Note:* You need to define the environment variables before installing the pods because there is a post install script which sets up the project with your environment variables. If you don't want to use environment variables, you can edit the `MiniApp-Secrets.xcconfig` created after a `pod install` in the parent folder of the project,but be aware you will have to update this file after every `pod install`
 
+<div id="test-sample-app"></div>
```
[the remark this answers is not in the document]
Climbatize: @corycaywood this is a typical web expert concern ^^

.travis.yml
```diff
@@ -42,6 +42,7 @@ after_success:
   --module MiniApp
   --source-directory MiniApp
   --podspec MiniApp.podspec
+  --theme fullwidth
```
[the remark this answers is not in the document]
ts-leojoseph-r: Done

DEV.md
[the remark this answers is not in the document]
ts-leojoseph-r: Done

ts-leojoseph-r:
> But I think there's still some work to be done on re-organizing the user guide. Specifically, I think we should make the structure more closely match Android ([see here](https://github.com/rakutentech/android-miniapp/blob/master/miniapp/USERGUIDE.md)). It doesn't have to exactly match, but I think it should be more similar.

I agree that it is good to have much similar documentation for both platforms.

> * Headers underneath "Getting Started" should be sub-headers - i.e. `## Requirements` -> `### Requirements`
> * Maybe the "Usage" section can be rearranged to more closely match the Android guide. i.e. rearrange it to have sub-sections like this:

Done

>   * "Create a Mini App"
>   * "Mini App Features" - contains info about communicating with mini app - `MiniAppMessageDelegate`, User Profile Details, Orientation Lock, etc.
>   * "Fetching Mini App Info" - contains "Load the MiniApp list:" and "Getting MiniAppInfo"
>   * "Custom Permissions"
>   * "Advanced Features" - anything else: ""Add a web navigation interface to the MiniApp view"", "Overriding config at runtime", "Allow full screen video", etc.

Done

> * I think some sections should have more detailed information on what the functionality does. See Android's "Create and display a Mini App" and  "Mini App Features" section, for example (maybe some of this text can just be copied directly?)

Done

> * iOS already has some direct links to API docs - but I also like the idea of adding the relevant API Docs at the top of each section (see Android). Maybe that would work here too?

API references are added now. But did you check the doc references in android. Looks like most of the API doc reference is broken

USERGUIDE.md
```diff
@@ -59,83 +52,44 @@ If you don't want to use project settings, you have to pass this information one
...
+### Create a MiniApp for the given `MiniAppId` :
 ---
-MiniApp library calls are done via the `MiniApp.shared()` singleton with or without a `MiniAppSdkConfig` instance (you can get the current one with `Config.getCurrent()`). If you don't provide a config instance, values in custom iOS target properties will be used by default. 
-
-```swift
-MiniApp.shared().list { (result) in
-	...
-}
-```
+**API Docs:** [MiniApp.create](https://rakutentech.github.io/ios-miniapp/Classes/MiniApp.html), [MiniAppDisplayDelegate](https://rakutentech.github.io/ios-miniapp/Protocols/MiniAppDisplayDelegate.html)
```
corycaywood, line 79: In the case of iOS, you don't need to manually add the URL: Jazzy can generate a link automatically if you put a class name or method reference in backticks. For example, this should work:
```suggestion
**API Docs:** `MiniApp.create(appId:completionHandler:messageInterface:)`, `MiniAppDisplayDelegate`
```

See the Jazzy docs:
https://github.com/realm/jazzy#supported-documentation-keywords

USERGUIDE.md
```diff
@@ -221,153 +222,210 @@ extension ViewController: MiniAppMessageDelegate {
...
-```swift
-let navConfig = MiniAppNavigationConfig(
-                    navigationBarVisibility: .always,
-                    navigationDelegate: myNavigationDelegate,
-                    customNavigationView: mCustomView)
+<a id="user-profile-details-username"></a>
 
-MiniApp.shared(with: Config.getCurrent(), navigationSettings: navConfig).info(miniAppId: miniAppID) { (result) in
-...
+#### User Name
```
corycaywood, line 236: I think this should be `######` so it's smaller than `Retrieve User Profile details`:

```suggestion
###### User Name
```

USERGUIDE.md
```diff
@@ -221,153 +222,210 @@ extension ViewController: MiniAppMessageDelegate {
...
     }
-
-    ...
 }
 ```
 
-The `externalLinkResponseHandler` closure allows you to give a feedback as an URL to the SDK, for example when the controller is closed or when a custom scheme link is tapped. This closure can be passed to a `MiniAppExternalUrlLoader` object that will provide a method to test an URL and return the appropriate decision for a `WKNavigationDelegate` method, and if you provided a controller it will be dismissed automatically. Here is an example following previous example:
+<a id="access-token-info"></a>
+
+#### Access Token Info
```
corycaywood, line 266:
```suggestion
###### Access Token Info
```

USERGUIDE.md
```diff
@@ -221,153 +222,210 @@ extension ViewController: MiniAppMessageDelegate {
...
+        return ""
+    }
 }
 ```
 
-#### Opening external links
+<a id="user-profile-details-profilephoto"></a>
 
-By default MiniApp iOS SDK will open external links into a separate modal controller when tapped. `MiniAppNavigationDelegate` implements a method that allows to override this behaviour and provide your own external links management. Here is an example of implementation:
+#### Profile Photo
```
corycaywood, line 251:
```suggestion
###### Profile Photo
```

USERGUIDE.md
```diff
@@ -59,83 +52,44 @@ If you don't want to use project settings, you have to pass this information one
...
-
-```swift
-MiniApp.shared().list { (result) in
-	...
-}
-```
+**API Docs:** [MiniApp.create](https://rakutentech.github.io/ios-miniapp/Classes/MiniApp.html), [MiniAppDisplayDelegate](https://rakutentech.github.io/ios-miniapp/Protocols/MiniAppDisplayDelegate.html)
 
-or
+`MiniApp.create` is used to create a `View` for displaying a specific mini app. You must provide the mini app ID which you wish to create (you can get the mini-app ID by [Fetching Mini App Info](#fetching-mini-app-info) first). Calling `MiniApp.create` will do the following:
```
corycaywood, line 81: Looks like this link isn't working. I think it should be:

```suggestion
`MiniApp.create` is used to create a `View` for displaying a specific mini app. You must provide the mini app ID which you wish to create (you can get the mini-app ID by [Loading the Mini App List](#load-miniapp-list) first). Calling `MiniApp.create` will do the following:
```

corycaywood (approved):
Looks good! 👍  Just left a few minor comments.

corycaywood:
> API references are added now. But did you check the doc references in android. Looks like most of the API doc reference is broken

@leorjoseph Regarding this, the links in the Android documentation are relative to the published docs, so they will only work once published to the GH Pages site.

See my comment about Jazzy - you don't need to manually set the links, you can just have Jazzy generate them for you. Unfortunately, we aren't able to do something like this in the Android docs currently 😣

PR #186, opened 2020-12-09
Admob SDK integration

Tonbouy:
# Description
Implementation for Admob SDK. Still a draft PR as of now because there are still some grey areas, and things to solve, raising it now as a draft for visibility.

## Links
Associated ticket: https://jira.rakuten-it.com/jira/browse/MINIBUILD-518
Associated confluence page: https://confluence.rakuten-it.com/confluence/display/ONEAPP/Google+Ad+Mob+in+Mini+Apps


MiniApp.podspec
```diff
@@ -26,10 +26,15 @@ Pod::Spec.new do |s|
...
+  s.resource_bundle = {
+    "MiniApp" => ['MiniApp/**/*.{xcassets,js,pdf,xib}','js-miniapp/bridge.js'],
+    "Localization" => ["MiniApp/*.lproj/*.strings"]
+  }
+
+  s.static_framework = true
 
   s.dependency 'RSDKUtils', '>= 1.1.0'
   s.dependency 'ZIPFoundation'
+  s.dependency 'Google-Mobile-Ads-SDK'
```
Climbatize, line 38: Can we make a subspec for this, like `MiniApp/Ads` so developers who don't want to depend from Google Ads can. What do you think @corycaywood ?

Example/Info.plist
```diff
@@ -49,18 +49,27 @@
...
+	</dict>
+	<key>GADApplicationIdentifier</key>
+	<string>ca-app-pub-4264098964213781~1470114368</string>
+	<key>SKAdNetworkItems</key>
+	<array>
+		<dict>
+			<key>SKAdNetworkIdentifier</key>
+			<string>cstr6suwn9.skadnetwork</string>
+		</dict>
+	</array>
```
Climbatize, line 73: let's not commit IDs into master

MiniApp/Classes/MiniAppAdmobDisplayer.swift
```diff
@@ -0,0 +1,85 @@
...
+	private func createAndLoadRewarded() {
+		rewarded = GADRewardedAd(adUnitID: "ca-app-pub-3940256099942544/1712485313")
+		rewarded?.load(GADRequest())
+	}
+
+	private func createAndLoadInterstitial() {
+		interstitial = GADInterstitial(adUnitID: "ca-app-pub-3940256099942544/4411468910")
+		interstitial.delegate = self
+		interstitial.load(GADRequest())
+	}
```
Climbatize, line 30: let's extract these keys to some project config

MiniApp/Classes/MiniAppAdmobDisplayer.swift
```diff
@@ -0,0 +1,85 @@
...
+		if !isAdmobInitialized {
+			GADMobileAds.sharedInstance().start(completionHandler: {
+				[weak self] _ in
+
+				self?.isAdmobInitialized = true
+				self?.showInterstitial(onClosed: onClosed, onFailed: onFailed)
+			})
+		} else if interstitial.isReady {
+			self.onInterstitialClosed = onClosed
+			// PIERRE: here I need some reference to the viewController displaying the miniApp to allow the `present`
```
Climbatize, line 42: I understand. We usually take to top most presented controller to present another one. Typically:
```swift
        if #available(iOS 13, *) {
            window = UIApplication.shared.windows.filter { $0.isKeyWindow }.first
        } else {
            window = UIApplication.shared.keyWindow
        }
       [...]
        window?.topController()?.present(interstitial, animated: true)
```

Climbatize (commented):
Nice draft, let's discuss with the team about have to make it final

Example/Info.plist
corycaywood: The Admob App ID for the Example app should be set with an environment variable, similar to the other IDs in the App. See here:
https://github.com/rakutentech/ios-miniapp/blob/master/configure-secrets.sh

MiniApp.podspec
corycaywood: Yeah, I agree. Google Ads SDK should be an optional dependency because not every Host App will want to use Admob.

MiniApp/Classes/MiniAppAdmobDisplayer.swift
```diff
@@ -0,0 +1,89 @@
...
+		GADMobileAds.sharedInstance().start(completionHandler: { [weak self] _ in
+			self?.createAndLoadInterstitial()
+			self?.createAndLoadRewarded()
+		})
+
+	}
+
+	private func createAndLoadRewarded() {
+		//TEST: Google's test id from tutorial
+		rewarded = GADRewardedAd(adUnitID: "ca-app-pub-3940256099942544/1712485313")
```
corycaywood, line 25: Using a real id for testing can get your account suspended, so you should use Google's test ad unit IDs for testing: https://developers.google.com/admob/android/test-ads

Also, the ad unit IDs should be passed from the mini app (Javascript) side, so you don't need to have any IDs set on the native side. Our JS SDK Sample App is already set up to use test IDs, so you can use that for testing. See https://github.com/rakutentech/js-miniapp/blob/master/js-miniapp-sample/src/pages/ads.js#L103

MiniApp/Classes/MiniAppAdmobDisplayer.swift
```diff
@@ -0,0 +1,89 @@
...
+	var interstitial: GADInterstitial!
+	var rewarded: GADRewardedAd?
+	var onInterstitialClosed: (() -> Void)?
+	var onRewardedClosed: ((MiniAppReward?) -> Void)?
+	var lastReward: MiniAppReward?
+
+	override init() {
+		super.init()
+
+		GADMobileAds.sharedInstance().start(completionHandler: { [weak self] _ in
```
corycaywood, line 16: The initialization of the Google Ads SDK should be done by the Host App. In the SDK, we can't know whether the Host App has already initialized the Ads SDK or not.

MiniApp.xcodeproj/project.pbxproj
```diff
@@ -68,6 +68,8 @@
 		38F73B4523D554FE00F2C1CE /* Helpers.swift in Sources */ = {isa = PBXBuildFile; fileRef = 38F73B4423D554FE00F2C1CE /* Helpers.swift */; };
 		38F73B4723D55CCB00F2C1CE /* MiniAppInfoFetcherTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 38F73B4623D55CCB00F2C1CE /* MiniAppInfoFetcherTests.swift */; };
 		38FC4653244628F600BE33E0 /* MiniAppScriptMessageHandlerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 38FC4652244628F600BE33E0 /* MiniAppScriptMessageHandlerTests.swift */; };
+		3EF9F8CBAC6218D1B2F3D41A /* Pods_MiniApp_Tests.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 07EFB326C5FA5E5F73FE29B6 /* Pods_MiniApp_Tests.framework */; };
```
corycaywood, line 71: Please make sure to run `pod deintegrate` before committing.

corycaywood (commented):
Hi @Tonbouy , thanks for your contribution. It's looking like a good start!

I only reviewed this partially because you still need to integrate this with the JavaScript message bridge. See the examples here for integrating JS to iOS communication: 
https://github.com/rakutentech/ios-miniapp/blob/master/MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift

As a reminder, the mini app needs to be able to do two things with both interstitial and rewarded ads:
- Pre-Load any number of ad unit IDs which the mini app can specify at runtime (MiniApp.loadInterstitialAd & MiniApp.loadRewardedAd)
- Trigger a pre-loaded Ad to be displayed (MiniApp.showInterstitialAd & MiniApp.showRewardedAd)

You can test this functionality directly in our JS SDK Sample App. Probably the easiest way to do that is to use the Example App's "Load Mini App by URL" feature with this URL:
https://rakutentech.github.io/js-miniapp/sample

ts-leojoseph-r (commented):
Thank you for your contribution. I see it's still in draft version though, hence added only few comments. I believe you are still working on changes related to the Javascript bridge and UAT.

MiniApp/Classes/Models/MiniAppReward.swift
```diff
@@ -0,0 +1,6 @@
+import Foundation
+
+public struct MiniAppReward: Codable {
```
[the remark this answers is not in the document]
Tonbouy: Agreed, but for that one, even I don't really know what should be explained here. 

I just followed what was set as requirement and using it to return a response.. 🤔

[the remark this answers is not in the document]
ts-leojoseph-r: A simple one-liner description would do. For eg., `Struct to define Rewarded ads.` A similar one for the variables too. Leaving them undocumentated makes the developer documentation more [unclear like this](https://rakutentech.github.io/ios-miniapp/Enums/MASDKCustomPermissionError.html#/s:7MiniApp26MASDKCustomPermissionErrorO07unknownE0yA2CmF). We are trying to add as much as documentation now.

PR #187, opened 2020-12-09
Fix for Mini-2321 and Mini-2322

ts-leojoseph-r:
# Description
* Fix Custom permissions request UI 
*  Fix Location permissions error alert display

## Links

- MINI-2321
- MINI-2322

# Checklist

### DIFF
--- a/Example/Controllers/Extensions/ViewController+MiniAppMessageProtocol.swift
+++ b/Example/Controllers/Extensions/ViewController+MiniAppMessageProtocol.swift
@@ -33,7 +33,7 @@ extension ViewController: MiniAppMessageDelegate, CLLocationManagerDelegate {
     }
 
     func displayLocationDisabledAlert() {
-        DispatchQueue.main.async {
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
             let alert = UIAlertController(title: "Location Services are disabled", message: "Please enable Location Services in your Settings", preferredStyle: .alert)
             let okAction = UIAlertAction(title: "OK", style: .default, handler: nil)
             alert.addAction(okAction)
--- a/MiniApp/Classes/Display/ViewControllers/Custom Permissions/MACustomPermissionCell.xib
+++ b/MiniApp/Classes/Display/ViewControllers/Custom Permissions/MACustomPermissionCell.xib
@@ -17,19 +17,16 @@
                 <rect key="frame" x="0.0" y="0.0" width="414" height="75"/>
                 <autoresizingMask key="autoresizingMask"/>
                 <subviews>
-                    <switch opaque="NO" contentMode="scaleToFill" horizontalHuggingPriority="750" verticalHuggingPriority="750" contentHorizontalAlignment="center" contentVerticalAlignment="center" on="YES" translatesAutoresizingMaskIntoConstraints="NO" id="M4f-PF-F5r">
-                        <rect key="frame" x="352" y="10" width="51" height="14"/>
-                    </switch>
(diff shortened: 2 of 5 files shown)


Example/Controllers/Extensions/ViewController+MiniAppMessageProtocol.swift
```diff
@@ -33,7 +33,7 @@ extension ViewController: MiniAppMessageDelegate, CLLocationManagerDelegate {
     }
 
     func displayLocationDisabledAlert() {
-        DispatchQueue.main.async {
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
```
ts-leojoseph-r, line 36: Mini app requests location Custom permission and requests for location details as well. The sample app is trying to present both at the same time, hence the delay is added.

corycaywood: I don't really like using a timeout because you can never predict what the timeout should be for all devices/cases. But probably it's okay in this case since it's just the demo app.

However, I don't think this fully addresses the issues: it should be impossible for the mini app to access the location if the user denied the custom permission (without the Host App needing to do anything special). For example, when the Host App already has been granted the location device permission but the mini app was denied the location custom permission.

Looks like we need to add a check to `isUserAllowedPermssion` before accessing the location here: https://github.com/rakutentech/ios-miniapp/blob/4a18ace6594ad5129552abda26be3b82c4ceba4d/MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift#L56

ts-leojoseph-r: > I don't really like using a timeout because you can never predict what the timeout should be for all devices/cases. But probably it's okay in this case since it's just the demo app.

Me either.

> Looks like we need to add a check to `isUserAllowedPermssion` before accessing the location here:
> 
> https://github.com/rakutentech/ios-miniapp/blob/4a18ace6594ad5129552abda26be3b82c4ceba4d/MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift#L56

[Did you mean this one? ](https://github.com/rakutentech/ios-miniapp/pull/181/commits/32a04804187982c724a94d8de990e6800fd87176?file-filters%5B%5D=.swift#diff-fd17266a8049bc8366e4f1fba97d7a5bc0568eb62b4d3a2bf10122daefbe90deR83)

corycaywood: > > Looks like we need to add a check to `isUserAllowedPermssion` before accessing the location here:
> > https://github.com/rakutentech/ios-miniapp/blob/4a18ace6594ad5129552abda26be3b82c4ceba4d/MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift#L56
> 
> [Did you mean this one? ](https://github.com/rakutentech/ios-miniapp/pull/181/commits/32a04804187982c724a94d8de990e6800fd87176?file-filters%5B%5D=.swift#diff-fd17266a8049bc8366e4f1fba97d7a5bc0568eb62b4d3a2bf10122daefbe90deR83)

But that's only checked when the mini app calls `requestPermission`, right? We need to also check `isUserAlledPermission` when the mini app calls `getCurrentPosition`.

For example, try this:
1. In the JS SDK Sample App - "Turn On" device location
2. Accept both the custom permission and device permission
3. In Permission settings screen, disable location permission for sample app
4. In Sample App, "Turn On" device location again - but this time deny the custom location permission

In this case, the mini app is still able to receive the location - it should instead receive an error that location permission was denied.

ts-leojoseph-r: Done

corycaywood (approved):
Looks good 👍

Climbatize (approved):
LGTM

PR #188, opened 2020-12-10
feature: get contacts

minh-rakuten:
# Description
Allow miniapp to get contact list from hostapp.
Can be tested with JS Sample app version 1.6.

## Links
MINI-2296

# Checklist

### DIFF
--- a/Example/Controllers/Extensions/ViewController+MiniAppUserInfoProtocol.swift
+++ b/Example/Controllers/Extensions/ViewController+MiniAppUserInfoProtocol.swift
@@ -20,7 +20,7 @@ extension ViewController {
         guard let userProfile = getProfileSettings(), let contactList = userProfile.contactList else {
             return nil
         }
-        return contactList.map { MAContact(id: $0.id) }
+        return contactList
     }
 
     func getAccessToken(miniAppId: String, completionHandler: @escaping (Result<MATokenInfo, MASDKCustomPermissionError>) -> Void) {
--- a/Example/Controllers/Extensions/ViewController+MiniAppUserInfoProtocol.swift
+++ b/Example/Controllers/Extensions/ViewController+MiniAppUserInfoProtocol.swift
@@ -20,8 +20,7 @@ extension ViewController {
         guard let userProfile = getProfileSettings(), let contactList = userProfile.contactList else {
             return nil
         }
-        let contacts = contactList.map { MAContact(id: $0.id) }
-        return contacts
+        return contactList.map { MAContact(id: $0.id) }
     }
 
     func getAccessToken(miniAppId: String, completionHandler: @escaping (Result<MATokenInfo, MASDKCustomPermissionError>) -> Void) {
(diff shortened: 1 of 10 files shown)


Example/Controllers/Extensions/ViewController+MiniAppUserInfoProtocol.swift
```diff
@@ -16,6 +16,14 @@ extension ViewController {
         return userProfilePhoto
     }
 
+    func getContacts() -> [MAContact]? {
+        guard let userProfile = getProfileSettings(), let contactList = userProfile.contactList else {
+            return nil
+        }
+        let contacts = contactList.map { MAContact(id: $0.id) }
```
ts-leojoseph-r, line 23: You can return them just by `return contactList.map { MAContact(id: $0.id) }`

Example/Models/UserProfileModel.swift
```diff
@@ -3,15 +3,23 @@ import UIKit
...
 }
 
+struct Contact: Codable {
+    var id: String
+
+    init(id: String) {
+        self.id = id
+    }
+}
+
```
ts-leojoseph-r, line 22: We can use the public class `MAContact` right ? This seems a duplicate just for the sample app

MiniApp/Classes/Protocols/MiniAppUserInfoDelegate.swift
```diff
@@ -37,6 +44,14 @@ public class MATokenInfo: Codable {
...
 }
 
+public class MAContact: Codable {
+    let id: String
+
+    public init(id: String) {
+        self.id = id
+    }
+}
+
```
ts-leojoseph-r, line 54: Some developer documentation maybe ?

ts-leojoseph-r (commented):
Thanks for the PR @minh-rakuten . Too bad that I worked on this change already in local was thinking to push it tomorrow 😆 . Never mind. Added few comments, have a look at them.

Example/Models/UserProfileModel.swift
minh-rakuten: I don't know why but I cannot import `MAContact`, neither of `MATokenInfo`.

corycaywood: I think this should work with `MAContact` - it was working for me when I tried it. 
@minh-rakuten Did you try cleaning the XCode build folder? Also make sure to add `import MiniApp` to this file.

MiniApp/Classes/Protocols/MiniAppUserInfoDelegate.swift
```diff
@@ -9,6 +9,9 @@ public protocol MiniAppUserInfoDelegate: class {
     /// Interface that is used to retrieve the Image URI
     func getProfilePhoto() -> String?
 
+    /// Interface that is used to retrieve the Contact list
+    func getContacts() -> [MAContact]?
```
corycaywood, line 13: This one also needs to be added to the user guide.

corycaywood (commented):
Looking good 👍  

@minh-rakuten is an iOS developer now? 😄

Example/Models/UserProfileModel.swift
minh-rakuten: Thanks. Re-build in XCode works for me.

ts-leojoseph-r (approved):
Looks good to me :+1: 👏

corycaywood (approved):
Looks good! 👍

PR #191, opened 2020-12-17
Update User profile interfaces to async

ts-leojoseph-r:
# Description
Update the following interfaces into async

- getUserName()
- GetProfilePhoto()

## Links
MINI-2121

# Checklist

### DIFF
--- a/MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift
+++ b/MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift
@@ -176,11 +176,7 @@ internal class MiniAppScriptMessageHandler: NSObject, WKScriptMessageHandler {
                     }
                     self.executeJavaScriptCallback(responseStatus: .onSuccess, messageId: callbackId, response: userName)
                 case .failure(let error):
-                    if !error.localizedDescription.isEmpty {
-                        self.executeJavaScriptCallback(responseStatus: .onError, messageId: callbackId, response: error.localizedDescription)
-                        return
-                    }
-                    self.executeJavaScriptCallback(responseStatus: .onError, messageId: callbackId, response: getMiniAppErrorMessage(MiniAppErrorType.unknownError))
+                    self.handleMASDKError(error: error, callbackId: callbackId)
                 }
             }
         } else {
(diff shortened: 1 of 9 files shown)


Climbatize (approved):
Looks good to me. I just have one suggestion. You could keep the old methods without deprecating them. Here is a suggestion you could apply to all your methods to keep retro-compatibility:

keep the delegate as is and implement it in host app (you can even simplify the one that was in place):
```swift
func getUserName() -> String? {
        return getProfileSettings()?.displayName
}
```
and in the public extension of MiniAppUserInfoDelegate as a default implementation for the new method make a call to the old one:

```swift
func getUserName(completionHandler: @escaping (Result<String?, MASDKError>) -> Void?) {
        guard let userName = getUserName() else {
            return completionHandler?(.failure(.unknownError(domain: "Unknown Error", code: 1, description: "Failed to retrieve User name")))
        }
        completionHandler(.success(userName)) ?? username
    }
```

ts-leojoseph-r:
> Looks good to me. I just have one suggestion. You could keep the old methods without deprecating them. Here is a suggestion you could apply to all your methods to keep retro-compatibility:
> 
> keep the delegate as is and implement it in host app (you can even simplify the one that was in place):

Thanks for the suggestion, but it is the Host app that requested to make this Async, keeping the interface the same and just changing the default extension wouldn't be a solution that they would need.

MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift
```diff
@@ -167,26 +167,48 @@ internal class MiniAppScriptMessageHandler: NSObject, WKScriptMessageHandler {
...
+            hostAppMessageDelegate?.getProfilePhoto { (result) in
+                switch result {
+                case .success(let response):
+                    guard let profilePhoto = response else {
+                        self.executeJavaScriptCallback(responseStatus: .onError, messageId: callbackId, response: getMiniAppErrorMessage(MiniAppErrorType.hostAppError))
+                        return
+                    }
+                    self.executeJavaScriptCallback(responseStatus: .onSuccess, messageId: callbackId, response: profilePhoto)
+                case .failure(let error):
+                    if !error.localizedDescription.isEmpty {
```
corycaywood, line 202: It seems like there's some code duplication with the `.failure` case. Most of the functions in this class do the same thing on failure. Is there a way to extract this into another function?

corycaywood (approved):
Looks good 👍  Just left a suggestion.

MiniApp/Classes/JavascriptBridge/MiniAppScriptMessageHandler.swift
ts-leojoseph-r: Done

PR #199, opened 2021-01-05
Feat/query params

ts-leojoseph-r:
# Description
* Changes that allows the Host app to pass optional `String` parameters while creating/launching the miniapp.
* Changes in Sample app settings page to add the optional `String` parameters under `General` section

## Links
Mini-2296

# Checklist


MiniApp/Classes/Display/MiniAppWebView.swift
```diff
@@ -12,21 +12,31 @@ internal class MiniAppWebView: WKWebView {
...
         let urlRequest = URLRequest(url: miniAppURL)
         self.init(frame: .zero, configuration: MiniAppWebView.defaultConfig())
         commonInit(urlRequest: urlRequest)
     }
 
+    private static func getQueryParams(queryParams: String?) -> String {
+        guard let queryString = queryParams else {
+            return ""
+        }
+        if queryString.hasPrefix("?") {
```
khairul-alam-licon, line 34: In Android side, we haven't this kind of checking, or, any other additional checking for now. 
HostApp can pass `?demoParms` or, `??demoParms` or even, `???demoParms`. So, maybe we can just return it by appending at-least one qus-mark. Therefore, `return "?" + queryString` is enough.

And, let MiniApp developers to parse in their side.

ts-leojoseph-r: I don't think query parameters would have multiple question marks `?`. 
Check https://tools.ietf.org/html/rfc3986#section-3.4

@corycaywood , whats your thoughts on this ?

MiniApp/Classes/Display/MiniAppWebView.swift
```diff
@@ -12,21 +12,31 @@ internal class MiniAppWebView: WKWebView {
         return config
     }
 
-    convenience init(miniAppId: String, versionId: String) {
+    convenience init(miniAppId: String, versionId: String, queryParams: String? = nil) {
         let schemeName = Constants.miniAppSchemePrefix + miniAppId
-        let urlRequest = URLRequest(url: URL(string: schemeName + "://miniapp/" + Constants.rootFileName)!)
+        let urlRequest = URLRequest(url: URL(string: schemeName + "://miniapp/" + Constants.rootFileName + MiniAppWebView.getQueryParams(queryParams: queryParams))!)
```
Climbatize, line 17: I've always thought we should not build the URL with concatenation. What do you think of a method executing this code:

```
        guard var urlComps = URLComponents(string: "miniapp/" + Constants.rootFileName) else {
            return nil
        }
        urlComps.scheme = schemeName
        urlComps.queryItems = "queryParams".components(separatedBy: "&").compactMap({ (param) -> URLQueryItem? in
            var components = param.components(separatedBy: "=")
            let name = components[0]
            let value: String
            if components.count > 1 {
                components.remove(at: 0)
                value = components.joined(separator: "=")
            }else {
                value = ""
            }
            return URLQueryItem(name: name, value: value)
        })
        
        guard let url = urlComps.url else {
            return nil
        }
        
        return URLRequest(url:url)
```

Climbatize (approved):
LGTM, let's think about my suggestion before merging it

MiniApp/Classes/Display/MiniAppWebView.swift
ts-leojoseph-r: Done

MiniApp/Classes/Display/MiniAppWebView.swift
```diff
@@ -12,21 +12,28 @@ internal class MiniAppWebView: WKWebView {
...
 
-    convenience init(miniAppURL: URL) {
+    convenience init(miniAppURL: URL, queryParams: String? = nil) {
         let urlRequest = URLRequest(url: miniAppURL)
         self.init(frame: .zero, configuration: MiniAppWebView.defaultConfig())
         commonInit(urlRequest: urlRequest)
     }
 
+    private static func getQueryParams(queryParams: String?) -> String {
+        guard let param = queryParams, let queryString =  param.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed) else {
```
corycaywood, line 31: I don't think we should use `addingPercentEncoding` - it should be the mini app's responsibility to ensure that they're passing/receiving encoded queries.

Also, they might want to use a URL fragment with `#` like this: `?param1=value#some_other_thing`. I think in this case `urlQueryAllowed` would encode the `#`, right?

corycaywood (commented):
Looking good, just one concern below.

MiniApp/Classes/Display/MiniAppWebView.swift
corycaywood: As mentioned below, I think we should make it the mini app's responsibility to parse the params. We should only care about two cases in the params:
- If the params can somehow break the loading of the mini app (prevent it from loading)
- If there is some security vulnerability with allowing certain types of params

MiniApp/Classes/Display/MiniAppWebView.swift
ts-leojoseph-r: Ah, yeah. Adding a percent-encoding is required here as the [URL](https://github.com/rakutentech/ios-miniapp/blob/master/MiniApp/Classes/Display/MiniAppWebView.swift#L17) will not accept a string with spaces, it will crash. To form a URL with space, we are adding this encoding.

w/ Encoding
`a b c` **->** `/index.html?a%20b%20c`
`?param1=value#some_other_thing` -> `/index.html??param1=value%23some_other_thing`

w/o encoding
`a b c` **->** CRASH
`?param1=value#some_other_thing` **->** `/index.html??param1=value#some_other_thing`

corycaywood: Ah, it's because on iOS you have to load the webview from a URLRequest instead of a plain String?

But it's a problem that the `#` is getting converted to `%23`. Mini apps might want to use a Url fragment. Could you manually parse the `#something` portion first before encoding the query? Looks like there's a separate `urlFragmentAllowed` so you could encode that part separately: https://developer.apple.com/documentation/foundation/nscharacterset/1412537-urlfragmentallowed

ts-leojoseph-r: I found a better way now. Please check

MiniApp/Classes/Display/MiniAppWebView.swift
```diff
@@ -12,21 +12,28 @@ internal class MiniAppWebView: WKWebView {
...
         let urlRequest = URLRequest(url: miniAppURL)
         self.init(frame: .zero, configuration: MiniAppWebView.defaultConfig())
         commonInit(urlRequest: urlRequest)
     }
 
+    private static func getQueryParams(queryParams: String?) -> String {
+        guard let param = queryParams, let queryString =  param.encodeURLParam() else {
+            return ""
+        }
+        return "?" + queryString
```
corycaywood, line 34: On Android, I think that @khairul-alam-licon made it so the `?` mark is added only if the provided string doesn't already start with a `?`. This is because it's very unlikely that a mini app would purposely want an extra `?` at the start of the query.

Could you do something similar on iOS so that the behavior is matching across platforms?

ts-leojoseph-r: 😂😂😂
I did make those changes already in this PR, Check here https://github.com/rakutentech/ios-miniapp/pull/199/commits/e889acaf99a7b57d8870c003e6237926fc17a7e7#diff-b98abf2a7523c021d82b52ed01322d682e337911e1dd79400645ef0001694185R30

But @khairul-alam-licon said they are not doing like that in Android so I had remove it.

corycaywood: Oh, sorry I didn't realize it was removed on Android.

That's fine with me as long as it matches across platforms.

corycaywood (approved):
Looks good 👍

khairul-alam-licon (approved):
Looks good 👍

PR #200, opened 2021-01-14
Rename protocol `requestPermission` to `requestDevicePermission`

ts-leojoseph-r:
# Description

- Rename `MiniAppScriptMessageHandler.requestPermission` to `requestDevicePermission`
- Rename `MiniAppPermissionType` to `MiniAppDevicePermissionType` 

## Links
[MINI-2119](https://jira.rakuten-it.com/jira/browse/MINI-2119)

# Checklist

### DIFF
--- a/MiniApp/Classes/Protocols/MiniAppMessageDelegate.swift
+++ b/MiniApp/Classes/Protocols/MiniAppMessageDelegate.swift
@@ -62,6 +62,10 @@ public extension MiniAppMessageDelegate {
     func requestPermission(permissionType: MiniAppDevicePermissionType, completionHandler: @escaping (Result<MASDKPermissionResponse, MASDKPermissionError>) -> Void) {
         requestDevicePermission(permissionType: permissionType, completionHandler: completionHandler)
     }
+
+    func requestDevicePermission(permissionType: MiniAppDevicePermissionType, completionHandler: @escaping (Result<MASDKPermissionResponse, MASDKPermissionError>) -> Void) {
+        completionHandler(.failure(.failedToConformToProtocol))
+    }
 }
 
 public enum MASDKProtocolResponse: String {
--- a/MiniApp/Classes/Protocols/MiniAppMessageDelegate.swift
+++ b/MiniApp/Classes/Protocols/MiniAppMessageDelegate.swift
@@ -60,11 +60,11 @@ public extension MiniAppMessageDelegate {
     }
 
     func requestPermission(permissionType: MiniAppDevicePermissionType, completionHandler: @escaping (Result<MASDKPermissionResponse, MASDKPermissionError>) -> Void) {
-        requestDevicePermission(permissionType: permissionType, completionHandler: completionHandler)
+        completionHandler(.failure(.failedToConformToProtocol))
     }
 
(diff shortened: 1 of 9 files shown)


MiniApp/Classes/Protocols/MiniAppMessageDelegate.swift
```diff
@@ -51,6 +58,10 @@ public extension MiniAppMessageDelegate {
             Result<[MASDKCustomPermissionModel], MASDKCustomPermissionError>) -> Void) {
         self.requestCustomPermissions(permissions: permissions, miniAppTitle: "Mini App", completionHandler: completionHandler)
     }
+
+    func requestPermission(permissionType: MiniAppDevicePermissionType, completionHandler: @escaping (Result<MASDKPermissionResponse, MASDKPermissionError>) -> Void) {
```
corycaywood, line 62: I think that you need to also add a default implementation for `requestDevicePermission` which will just call `requestPermission` (the reverse of this one).

The problem is that currently adding `requestDevicePermission` is a breaking change because the Host App will be required to add it. We still need to support Host Apps that have added `requestPermission` but not yet `requestDevicePermission`.

Do you think my suggestion would work for this?

corycaywood (commented):
Looking good, just one concern about backwards compatibility.

MiniApp/Classes/Protocols/MiniAppMessageDelegate.swift
ts-leojoseph-r: Ok. So in that case,  either `requestDevicePermission` or `requestPermission` interface will not be mandated. Is that ok?

corycaywood: @ts-leojoseph-r Yeah, I think that's okay since it's the only way to avoid a breaking change for the Host App. I think Android did something similar to this too.

ts-leojoseph-r: If we add the default implementation of `requestDevicePermission` which will call `requestPermission`. Then it will be loop right?. Because we already have default implementation of `requestPermission` which calls `requestDevicePermission`. 

So either we can have any one default implementation, or if we want both default implementation, then `requestDevicePermission` implementation should have this code https://github.com/rakutentech/ios-miniapp/blob/master/Example/Controllers/Extensions/ViewController%2BMiniAppMessageProtocol.swift#L7

corycaywood: @ts-leojoseph-r Ah, yeah you're right it would be a loop if they didn't implement one of them. On Android, this was actually solved differently - the default implementation for both `requestPermission` and `requestDevicePermission` is to throw an error to the Mini App. In the case of `requestDevicePermission`, it catches the error and tries to call `requestPermission`. So if neither one is implemented, then the mini app will receive an error message. Is something like that possible for iOS?

Actually, we did eventually want to have a default implementation for both of these, but we just haven't done it yet. So probably whichever one you think is easier would be okay.

corycaywood (approved):
LGTM 👍

Climbatize (approved):
LGTM

PR #207, opened 2021-01-22
Ads SDK integration

Climbatize:
# Description
based on #186 
Added Ads subspec to make Google ads framework an optional dependency.
Made the library static due to Google ads framework limitations
Made the library only compatible with iOS due to Google ads framework limitations

# Checklist


MiniApp/Classes/core/Display/MiniAppAdDisplayer.swift
```diff
@@ -0,0 +1,61 @@
...
+    override init() {
+        super.init()
+        initFramework()
+    }
+
+    func initFramework() {
+    }
+
+    func loadRequestedAd(forParams params: RequestParameters?, onLoaded: @escaping (Result<Void, Error>) -> Void) {
+        guard let delegate = delegate, let params = params, let adTypeRaw = params.adType, let adType = MiniAppAdType(rawValue: adTypeRaw), let adId = params.adUnitId else {
```
corycaywood, line 26: I think we should have a separate error message for the case that the host app didn't implement the delegate: i.e. the host app doesn't support displaying ads.

MiniApp/Classes/core/Display/MiniAppAdDisplayer.swift
```diff
@@ -0,0 +1,61 @@
...
+    func cleanInterstitial(_ adId: String) {
+        onInterstitialClosed.removeValue(forKey: adId)
+    }
+
+    func loadInterstitial(for adId: String, onLoaded: @escaping (Result<Void, Error>) -> Void) {
+        return onLoaded(.failure(NSError.miniAppAdNotLoaded(message: MASDKAdsDisplayError.failedToConformToProtocol.localizedDescription)))
+    }
+
+    func showInterstitial(for adId: String, onClosed: @escaping (Result<Void, Error>) -> Void) {
+        return onClosed(.failure(NSError.miniAppAdNotLoaded(message: MASDKAdsDisplayError.failedToConformToProtocol.localizedDescription)))
```
corycaywood, line 51: Looks like `showInterstitial` and `showRewarded` are missing the check for `delegate`. They should also check this and, if not implemented, then return some error message about the host app not supporting ads.

MiniApp/Classes/admob/AdMobDisplayer.swift
```diff
@@ -0,0 +1,153 @@
...
+    var onInterstitialLoaded: [String: (Result<Void, Error>) -> Void] = [:]
+    var admobStartResult: [String : GADAdapterStatus]?
+
+    override init() {
+        super.init()
+        delegate = self
+    }
+
+    override func initFramework() {
+        GADMobileAds.sharedInstance().start {[weak self] status in
```
corycaywood, line 17: I think it should be the host app's responsibility to initialize the Ad Mob SDK. For example, they might also be using Ad Mob somewhere else in their App so they would want to initialize it at App launch.

MiniApp/Classes/core/Display/MiniAppAdDisplayer.swift
```diff
@@ -0,0 +1,61 @@
+import Foundation
+
+/// This class is the parent class meant to be overriden when implementing ads SDK
+/// - Tag: MiniAppAdDisplayer
+internal class MiniAppAdDisplayer: NSObject, MiniAppAdDisplayDelegate {
+    #if RMA_SDK_ADMOB
+    static let shared = AdMobDisplayer()
```
corycaywood, line 7: I'm not sure that having a shared `MiniAppAdDisplayer` will work the way that we need.

The host app needs to declare the `MiniAppAdDisplayer` for each individual mini app. This is because some mini apps will use Ad Mob and some will use something else. So by default, a mini app should be launched without support for Ads, and it only supports Ads if a `MiniAppAdDisplayer` is supplied to `MiniApp.create`.

corycaywood (commented):
Looking good! I left a few comments below.

The only major thing is a small change to the default Ads behavior: by default, mini apps shouldn't support Ads (even when the AdMob sub-spec is included). The host app needs to provide the ads displayer to mini app separately when creating each mini app.

USERGUIDE.md
```diff
@@ -224,6 +233,53 @@ extension ViewController: MiniAppMessageDelegate {
...
+- by implementing [MiniAppAdDisplayDelegate](https://rakutentech.github.io/ios-miniapp/Protocols/MiniAppAdDisplayDelegate.html) by yourself 
+- if you rely on Google ads to display your ads you can simply implement `pod MiniApp/Admob` into your pod dependencies (see [settings section](#setting-admob)).
+
+###### Custom ads displayer
+
+If you chose to implement ads displaying by yourself, you first need do implement [MiniAppAdDisplayDelegate](https://rakutentech.github.io/ios-miniapp/Protocols/MiniAppAdDisplayDelegate.html) in a controller:
+
+```swift
+extension ViewController: MiniAppAdDisplayDelegate {
+    func loadInterstitial(forId: String) {
```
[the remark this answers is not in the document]
Climbatize: oh you're right, I forgot to update documentation today, thanks for the head up

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 **SDK**
 
 - **Feature** Added analytics for Mini App usage tracking
+- **Feature** Added ads for Mini Apps host
+- **Feature** Added Google mobile ads integration in a `MiniApps/Admob` pod subspec
```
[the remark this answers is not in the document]
Climbatize: sure

corycaywood (approved):
Looks good! 👍

MiniApp/Classes/core/Extensions/NSError+MiniApp.swift
```diff
@@ -14,54 +14,78 @@ extension NSError {
...
+
+    class func miniAppAdNotDisplayed(message: String) -> NSError {
+        NSError(
+                domain: MiniAppSDKErrorDomain,
+                code: MiniAppSDKErrorCode.adNotDisplayed.rawValue,
+                userInfo: [NSLocalizedDescriptionKey: message]
+        )
+    }
+
+    class func miniAppAdProtocoleError() -> NSError {
```
ts-leojoseph-r, line 82: miniAppAdProtocoleError -> miniAppAdProtocolError

MiniApp/Classes/core/Display/MiniAppAdDisplayer.swift
```diff
@@ -0,0 +1,60 @@
...
+
+    public func loadRewarded(for adId: String, onLoaded: @escaping (Result<Void, Error>) -> Void) {
+        if let delegate = delegate {
+            delegate.loadRewarded(for: adId, onLoaded: onLoaded)
+        } else {
+            return onLoaded(.failure(NSError.miniAppAdProtocoleError()))
+        }
+    }
+
+    public func showRewarded(for adId: String, onClosed: @escaping (Result<MiniAppReward, Error>) -> Void) {
```
ts-leojoseph-r, line 53: Could you please add some dev documentation for the public interfaces ?

MiniApp/Classes/core/JavascriptBridge/MiniAppScriptMessageHandler.swift
```diff
@@ -70,6 +72,36 @@ internal class MiniAppScriptMessageHandler: NSObject, WKScriptMessageHandler {
...
             fetchTokenDetails(callbackId: callbackId)
+        case .loadAd:
+            loadRequestedAd(forParams: requestParam) { [weak self] result in
+                switch result {
+                case .success:
+                    self?.executeJavaScriptCallback(responseStatus: .onSuccess, messageId: callbackId, response: "Ad loaded")
+                case .failure(let error):
+                    self?.executeJavaScriptCallback(responseStatus: .onError, messageId: callbackId, response: error.localizedDescription)
+                }
+            }
```
ts-leojoseph-r, line 83: Can we move this to separate method ?

MiniApp/Classes/core/Protocols/MiniAppAdDisplayDelegate.swift
```diff
@@ -0,0 +1,69 @@
...
+		case .sdkError:
+			return "Third party Ads SDK failed to provide an ad"
+		case .adNotLoaded:
+			return "Ad is not ready to launch"
+		case .rewardFailure:
+			return "A reward could not be generated"
+		case .hostUIError:
+			return "The host controller is not able to display the ad"
+		case .adIdError:
+			return "Ad identifiers not provided or wrong request parameters format"
```
ts-leojoseph-r, line 66: Should we make these Localizable? What do you think?

ts-leojoseph-r (commented):
Kudos, great work 👏  . Overall looks great to me 👍 . Just a few comments have a look at it.

MiniApp/Classes/core/Extensions/NSError+MiniApp.swift
Climbatize: I can count on you to fix my Frenchness ^^

MiniApp/Classes/core/Display/MiniAppAdDisplayer.swift
Climbatize: these are just protocol implementation, I already commented protocol definition, but I think it doesn't hurt to duplicate them

MiniApp/Classes/core/JavascriptBridge/MiniAppScriptMessageHandler.swift
Climbatize: It's already a method, but I see what you mean, I'll make this clearer

MiniApp/Classes/core/Protocols/MiniAppAdDisplayDelegate.swift
Climbatize: I think, mister, this is a good idea

Climbatize: Another strings for other features need to be localized, I'll submit a separate PR

ts-leojoseph-r (approved):
Looks good to me :+1:

Doc.podspec
```diff
@@ -0,0 +1,42 @@
+Pod::Spec.new do |s|
```
corycaywood, line 1: Why does Jazzy need a separate PodSpec?

It seems it will be difficult to maintain this because we have to keep two podspecs in sync. Is there any way to keep the config in just one podspec?

PR #212, opened 2021-01-28
Candidate: fix GPS location update

Climbatize:
# Description
Updated LocationManager to retrieve location update more efficiently

# Checklist
- [] I wrote/updated tests for new/changed code

### DIFF
--- a/MiniApp/Classes/core/JavascriptBridge/MiniAppScriptMessageHandler.swift
+++ b/MiniApp/Classes/core/JavascriptBridge/MiniAppScriptMessageHandler.swift
@@ -55,13 +55,7 @@ internal class MiniAppScriptMessageHandler: NSObject, WKScriptMessageHandler {
             requestDevicePermission(requestParam: requestParam, callbackId: callbackId)
         case .getCurrentPosition:
             locationManager = LocationManager(enableHighAccuracy: requestParam?.locationOptions?.enableHighAccuracy ?? false)
-            locationManager?.updateLocation {[weak self] result in
-                switch result {
-                case .success:
-                    self?.getCurrentPosition(callbackId: callbackId)
-                default: self?.executeJavaScriptCallback(responseStatus: .onError, messageId: callbackId, response: getMiniAppErrorMessage(MASDKCustomPermissionError.userDenied))
-                }
-            }
+            updateLocation(callbackId: callbackId)
         case .requestCustomPermissions:
             requestCustomPermissions(requestParam: requestParam, callbackId: callbackId)
         case .shareInfo:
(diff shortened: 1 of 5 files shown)


MiniApp/Classes/core/JavascriptBridge/MiniAppScriptMessageHandler.swift
```diff
@@ -269,11 +275,29 @@ internal class MiniAppScriptMessageHandler: NSObject, WKScriptMessageHandler {
...
         manager.desiredAccuracy = enableHighAccuracy ? kCLLocationAccuracyBest : kCLLocationAccuracyHundredMeters
+        manager.delegate = self
+    }
+
+    func updateLocation(result: @escaping (Result<CLLocation?, Error>) -> Void) {
+        if CLLocationManager.authorizationStatus() == .authorizedAlways || CLLocationManager.authorizationStatus() == .authorizedWhenInUse {
+            self.locationListener = result
+            manager.startUpdatingLocation()
+        } else {
+            result(.success(manager.location))
```
Climbatize, line 293: I kept this as a success because unit test failures proved that this was the expected behavior. I'd prefer returning a failure, but we need to think about it before

MiniApp/Classes/core/JavascriptBridge/MiniAppScriptMessageHandler.swift
```diff
@@ -55,7 +55,13 @@ internal class MiniAppScriptMessageHandler: NSObject, WKScriptMessageHandler {
...
         case .getCurrentPosition:
             locationManager = LocationManager(enableHighAccuracy: requestParam?.locationOptions?.enableHighAccuracy ?? false)
-            getCurrentPosition(callbackId: callbackId)
+            locationManager?.updateLocation {[weak self] result in
+                switch result {
+                case .success:
+                    self?.getCurrentPosition(callbackId: callbackId)
+                default: self?.executeJavaScriptCallback(responseStatus: .onError, messageId: callbackId, response: getMiniAppErrorMessage(MASDKCustomPermissionError.userDenied))
+                }
+            }
```
ts-leojoseph-r, line 64: Can we move this to a separate function?

MiniApp/Classes/core/JavascriptBridge/MiniAppScriptMessageHandler.swift
```diff
@@ -269,11 +275,29 @@ internal class MiniAppScriptMessageHandler: NSObject, WKScriptMessageHandler {
...
+        } else {
+            result(.success(manager.location))
+        }
+    }
+}
+
+extension LocationManager: CLLocationManagerDelegate {
+    func locationManager(_ manager: CLLocationManager, didUpdateLocations locations: [CLLocation]) {
+        locationListener?(.success(locations[0]))
+        manager.stopUpdatingLocation()
```
ts-leojoseph-r, line 301: ✌️

ts-leojoseph-r (approved):
Looks good to me :+1: Just one refactoring comment.

MiniApp/Classes/core/JavascriptBridge/MiniAppScriptMessageHandler.swift
Climbatize: of course

MiniApp/Classes/core/JavascriptBridge/MiniAppScriptMessageHandler.swift
corycaywood: I think this is the wrong behavior so we should change this here. On Android, if the user tries to retrieve the location while the permission is denied, then the Android OS sends an error like: 
```
{code: 2, message: "application does not have sufficient geolocation permissions."}
```

It actually returns an error object - but I think we don't have that capability right now in the JS Bridge. So probably it's okay to just return the error message `application does not have sufficient geolocation permissions.`.

corycaywood (commented):
Looking good 👍 But see my comment below.

Also, could you add this fix to the change log?

corycaywood (approved):
Looks good 👍

PR #213, opened 2021-02-01
Release 2.8.0

Climbatize:
merge release candidate into master


ts-leojoseph-r (commented):
I see lot of changes related to refactoring and location. Can we avoid this in Release PR's ?

Climbatize:
> I see lot of changes related to refactoring and location. Can we avoid this in Release PR's ?

no worries, these are the fixes that resulted from regression tests. In fact this is just the merge of the released v2.8.0 SDK ;)

scripts/deploy-to-app-center.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
```
corycaywood, line 1: I think this isn't correct - `#!/bin/sh` is used for system boot scripts, isn't it?

When you create a new script in the Bitrise UI, it automatically uses `#!/usr/bin/env bash` for Bash scripts, so I think we should keep that.

Climbatize: you're right, this was a failed attempt to fix something wrong

corycaywood (approved):
👍

ts-leojoseph-r (approved):
Looks good to me :+1:

PR #220, opened 2021-02-08
Feat/meta data fetching

ts-leojoseph-r:
# Description
* Added getMiniAppManifest(miniAppId:miniAppVersion:) interface to retrieve MiniAppManifest from the platform
* Display First-time launch screen with the retrieved meta data information.

## Links
[MINI-2507](https://jira.rakuten-it.com/jira/browse/MINI-2507)

# Checklist


MiniApp/Classes/core/Models/MiniAppManifest.swift
```diff
@@ -0,0 +1,40 @@
...
+
+/// Mini-app meta data information
+public struct MiniAppManifest: Codable {
+
+    /// List of required permissions for a mini-app
+    public var requiredPermissions: [MASDKCustomPermissionModel]?
+    /// List of optional permissions for a mini-app
+    public var optionalPermissions: [MASDKCustomPermissionModel]?
+    /// Key-value pair data that is received from the endpoint
+    public var exampleHostAppMetaData: [String: String]?
```
corycaywood, line 33: Sorry, I think I made this kind of confusing. `exampleHostAppMetaData` shouldn't be part of our SDK - it's just an example key. 

The Host App can use any additional keys that they want in the manifest. So we need a way to provide the entire manifest response to them. i.e. have a `manifest` variable here which maps the top-level key/values of the whole manifest. For example, the host app can call:

```
// This returns the raw string value: {"exampleKey": "test"}
// It's the Host App's responsibility to parse it
miniAppManifest.manifest["exampleHostAppMetaData"]
```

MiniApp/Classes/core/MiniAppInfoFetcher.swift
```diff
@@ -41,4 +41,39 @@ internal class MiniAppInfoFetcher {
...
+            case .failure(let error):
+                return completionHandler(.failure(error))
+            }
+        }
+    }
+
+    func prepareMiniAppManifest(metaDataResponse: MetaDataResponse) -> MiniAppManifest {
+        var manifest = MiniAppManifest()
+        manifest.requiredPermissions = metaDataResponse.reqPermissions?.compactMap {
+            guard let name = $0.name, let reason = $0.reason else {
```
corycaywood, line 65: I think that `reason` should be optional. I know that it's required in the [W3C Spec](https://w3c.github.io/miniapp/specs/manifest/#dfn-reqpermissions), but I don't think we're actually going to use it. So I think it will cause a lot of confusion if we make it required.

I think we already made `permissionRequestDescription` optional, right?

MiniApp/Classes/core/MiniAppInfoFetcher.swift
```diff
@@ -41,4 +41,39 @@ internal class MiniAppInfoFetcher {
...
+                                                        metaDataResponse: decodeResponse)))
+                }
+                return completionHandler(.failure(NSError.invalidResponseData()))
+            case .failure(let error):
+                return completionHandler(.failure(error))
+            }
+        }
+    }
+
+    func prepareMiniAppManifest(metaDataResponse: MetaDataResponse) -> MiniAppManifest {
```
corycaywood, line 62: Couldn't this method be `private`?

MiniApp/Classes/core/MiniAppInfoFetcher.swift
```diff
@@ -41,4 +41,39 @@ internal class MiniAppInfoFetcher {
...
+                return nil
+            }
+            return MASDKCustomPermissionModel(permissionName: MiniAppCustomPermissionType(rawValue: name)!, isPermissionGranted: .allowed, permissionRequestDescription: reason)
+        }
+        manifest.optionalPermissions = metaDataResponse.optPermissions?.compactMap {
+            guard let name = $0.name, let reason = $0.reason else {
+                return nil
+            }
+            return MASDKCustomPermissionModel(permissionName: MiniAppCustomPermissionType(rawValue: name)!, isPermissionGranted: .allowed, permissionRequestDescription: reason)
+        }
```
corycaywood, line 75: This code block looks almost exactly the same as the one above. Could you reduce duplication by pulling this into a private function which converts an array of `MACustomPermissionsResponse` to an array of `MASDKCustomPermissionModel`?

MiniApp/Classes/core/RealMiniApp.swift
```diff
@@ -194,6 +194,25 @@ internal class RealMiniApp {
             }
         }
     }
+
+    func retrieveMiniAppMetaData(appId: String,
+                                 version: String? = nil,
```
corycaywood, line 199: Why is `version` optional here? Below it looks like we just call return `.failure` if version is `nil`, so couldn't we just make it a required param?

MiniApp/Classes/core/Storage/MiniAppStatus.swift
```diff
@@ -105,11 +104,11 @@ class MiniAppStatus {
         return storedPermissionsList ?? [:]
     }
 
-    /// Method to delete the existing custom permissions from the Keychain. If a mini app is downloaded for the first time, it is safe to remove the existing custom permission in the keychain.
-    /// - Parameter appId: Mini app ID
-    func validateCustomPermissions(appId: String) {
-        if getMiniAppInfo(appId: appId) == nil {
-            self.miniAppKeyStore.removeKey(for: appId)
+    /// Method to delete the existing custom permissions from the Keychain. Keystore should keep Custom Permissions for only downloaded mini-apps
```
corycaywood, line 107: Just noting here: We'll also need to change our implementation of how custom permissions are stored. Instead of storing deny/grant for all custom permissions, we should store only the ones in the manifest. i.e. the permissions listed in the settings screen should only list the permissions which are in the Mini App's manifest.

But probably that needs to be in a separate PR when you add caching for the manifest.

corycaywood (commented):
Looking good so far 👍  See my comments below.

MiniApp/Classes/core/Models/MiniAppManifest.swift
```diff
@@ -0,0 +1,40 @@
...
+        case name,
+             reason
+    }
+}
+
+/// Mini-app meta data information
+public struct MiniAppManifest: Codable {
+
+    /// List of required permissions for a mini-app
+    public var requiredPermissions: [MASDKCustomPermissionModel]?
```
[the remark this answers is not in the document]
ts-leojoseph-r: Done

MiniApp/Classes/core/MiniAppInfoFetcher.swift
ts-leojoseph-r: Done

MiniApp/Classes/core/MiniAppInfoFetcher.swift
ts-leojoseph-r: Done

MiniApp/Classes/core/MiniAppInfoFetcher.swift
ts-leojoseph-r: Done

MiniApp/Classes/core/RealMiniApp.swift
ts-leojoseph-r: When I started the ticket I was updating the `MiniApp.info` which needs the `version` to be optional. Now that we have separate interface to get the manifest, I changed it.

MiniApp/Classes/core/Storage/MiniAppStatus.swift
ts-leojoseph-r: Yeah, I thought the same. We need to store only the permissions which MiniApp entitled for. Will work on it in the new ticket

MiniApp/Classes/core/MiniAppInfoFetcher.swift
ts-leojoseph-r: The class is internal already. Do we need to make it private? Because this method is also used in Helpers

Climbatize: I think it's good to keep it internal

MiniApp/Classes/core/Models/MiniAppManifest.swift
ts-leojoseph-r: Sorry. I just added the variable there with the mock json for understanding. Once the endpoints are ready, I will change something like `manifest` or `extras`. 

variable name defined here will be static but it will retrieve any dynamic key from the response.

Example/Controllers/MAFirstLaunchController.swift
```diff
@@ -9,25 +9,44 @@ class MAFirstLaunchController: UIViewController {
...
     override func viewDidLoad() {
         super.viewDidLoad()
+        let podBundle: Bundle = Bundle.main
+        let nib = UINib(nibName: "CustomPermissionCell", bundle: podBundle)
+        self.tableView.register(nib, forCellReuseIdentifier: "FirstLaunchCustomPermissionCell")
         setupUI()
     }
 
+    override func viewWillAppear(_ animated: Bool) {
+        tableView.reloadData()
```
Climbatize, line 30: don't forget to call `super.viewWillAppear(animated)`

Example/Controllers/MAFirstLaunchController.swift
```diff
@@ -9,25 +9,44 @@ class MAFirstLaunchController: UIViewController {
...
+        permissionsCollections = (miniAppManifest?.requiredPermissions ?? []) + (miniAppManifest?.optionalPermissions ?? [])
+    }
+
+    override func viewDidLayoutSubviews() {
+        tableViewHeightConstraint.constant = tableView.contentSize.height
```
Climbatize, line 35: don't forget to call `super.viewDidLayoutSubviews()`

Example/Controllers/MAFirstLaunchController.swift
```diff
@@ -39,6 +58,80 @@ class MAFirstLaunchController: UIViewController {
     }
 }
 
+// MARK: - UITableViewControllerDelegate
+extension MAFirstLaunchController: UITableViewDelegate, UITableViewDataSource {
+
```
Climbatize, line 63: you could also implement there the didSelectRow(at indexpath) with something like 
```swift
{
    tableView.deselectRow(at: indexPath, animated: true)
    if let cell = tableView.cellForRow(at: indexPath) as? FirstLaunchCustomPermissionCell {
        cell.toggle.isOn = !cell.toggle.isOn
        permissionValueChanged(cell.toggle)
    }
}
```

Example/Controllers/MAFirstLaunchController.swift
```diff
@@ -39,6 +58,80 @@ class MAFirstLaunchController: UIViewController {
...
+        toggle.isOn = true
+        permissionTitle.text = ""
+        permissionDescription.text = ""
+    }
+}
+
+extension NSMutableAttributedString {
+    func highlightRedColor(_ value: String) -> NSMutableAttributedString {
+        let attributes: [NSAttributedString.Key: Any] = [
+            .foregroundColor: UIColor.red
```
Climbatize, line 127: if you wanted to use the crimson red you can access in `self.window?.tintColor` as it is defined in the app delegate

MiniApp/Classes/core/API/MetaDataAPI.swift
```diff
@@ -0,0 +1,22 @@
+internal class MetaDataAPI {
+    let environment: Environment
+
+    init(environment: Environment) {
```
Climbatize, line 4: not really important but maybe `init(with environment: Environment)` is more elegant

Climbatize (approved):
LGTM

Example/Controllers/MAFirstLaunchController.swift
ts-leojoseph-r: No, I was thinking the same at first but later I realized that's against the apple standard approach. Check any items in iOS Settings. User can toggle on/off only by tapping on the toggle button.

Example/Controllers/MAFirstLaunchController.swift
ts-leojoseph-r: Done

corycaywood (approved):
LGTM 👍  Just one comment below

MiniApp/Classes/core/Models/MiniAppManifest.swift
```diff
@@ -0,0 +1,57 @@
...
+
+/// Mini-app meta data information
+public struct MiniAppManifest: Codable {
+
+    /// List of required permissions for a mini-app
+    public let requiredPermissions: [MASDKCustomPermissionModel]?
+    /// List of optional permissions for a mini-app
+    public let optionalPermissions: [MASDKCustomPermissionModel]?
+    /// Key-value pair data that is received from the endpoint
+    public let exampleHostAppMetaData: [String: String]?
```
corycaywood, line 42: @leorjoseph This one needs to be updated to `customMetaData` to match Android.

MiniApp/Classes/core/Models/MiniAppManifest.swift
```diff
@@ -0,0 +1,57 @@
...
+        case bundleManifest
+    }
+}
+
+/// Mini App Meta data information
+internal struct MetaDataCustomPermissionModel: Decodable {
+    var reqPermissions: [MACustomPermissionsResponse]?
+    var optPermissions: [MACustomPermissionsResponse]?
+
+    var exampleHostAppMetaData: [String: String]?
```
corycaywood, line 15: Sam here, this should be `customMetaData`.

MiniApp/Classes/core/Models/MiniAppManifest.swift
ts-leojoseph-r: Done

MiniApp/Classes/core/Models/MiniAppManifest.swift
ts-leojoseph-r: Done

USERGUIDE.md
```diff
@@ -432,6 +433,19 @@ MiniApp.shared(with: Config.getCurrent()).info(miniAppId: miniAppID) { (result)
 }
 ```
 
+<a id="get-mini-meta-data"></a>
+
+### Getting a `MiniApp meta-data` :
```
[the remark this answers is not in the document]
ts-leojoseph-r: Done

PR #225, opened 2021-02-18
Refactor/remove rsdk utils

ts-leojoseph-r:
# Description
* Added more interfaces to the existing MiniAppKeyChain class that helps the MiniAppCacheVerifier to use the same.
* Removed the RSDK Pods as MiniAppKeyChain is a replacement.

# Checklist

### DIFF
--- a/MiniApp/Classes/core/Storage/MiniAppCacheVerifier.swift
+++ b/MiniApp/Classes/core/Storage/MiniAppCacheVerifier.swift
@@ -1,19 +1,18 @@
 import CommonCrypto
-import RSDKUtils
 
 internal class MiniAppCacheVerifier {
-    let keystore = KeyStore()
+    let keystore = MiniAppKeyChain(serviceName: .cacheVerifier)
 
     private func generateKeyId(for appId: String, version: String) -> String {
         "\(appId) | \(version)"
     }
 
     func verify(appId: String, version: String) -> Bool {
-        if keystore.key(for: appId) != nil {
+        if keystore.getCacheInfo(for: appId) != nil {
             return false // v2.5 legacy cleaning
         }
         let cachedFilesHash = calculateHash(appId: appId, version: version)
-        let storedHash = keystore.key(for: generateKeyId(for: appId, version: version)) ?? ""
+        let storedHash = keystore.getCacheInfo(for: generateKeyId(for: appId, version: version)) ?? ""
         return cachedFilesHash == storedHash
     }
 
@@ -21,7 +20,7 @@ internal class MiniAppCacheVerifier {
         keystore.removeKey(for: appId) // v2.5 legacy cleaning
         let appKey = generateKeyId(for: appId, version: version)
         keystore.removeKey(for: appKey)
-        keystore.addKey(key: calculateHash(appId: appId, version: version), for: appKey)
+        keystore.setCacheInfo(key: calculateHash(appId: appId, version: version), for: appKey)
     }
 
     private func calculateHash(appId: String, version: String) -> String {
--- a/MiniApp/Classes/core/Storage/MiniAppCacheVerifier.swift
+++ b/MiniApp/Classes/core/Storage/MiniAppCacheVerifier.swift
@@ -1,26 +1,26 @@
 import CommonCrypto
 
 internal class MiniAppCacheVerifier {
-    let keystore = MiniAppVerificationStorage()
+    let miniAppVerificationStore = MiniAppVerificationStorage()
 
     private func generateKeyId(for appId: String, version: String) -> String {
         "\(appId) | \(version)"
     }
 
     func verify(appId: String, version: String) -> Bool {
-        if keystore.getCacheInfo(for: appId) != nil {
+        if miniAppVerificationStore.getCacheVerificationInfo(for: appId) != nil {
             return false // v2.5 legacy cleaning
         }
         let cachedFilesHash = calculateHash(appId: appId, version: version)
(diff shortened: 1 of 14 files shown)


MiniApp/Classes/core/Storage/MiniAppKeyChain.swift
```diff
@@ -74,11 +76,78 @@ import Foundation
...
 
+        guard let keys = try? JSONSerialization.jsonObject(with: storedData, options: []) as? CacheVerifierKeysDictionary else {
+            return nil
+        }
+
+        return keys
+    }
+
+    private func writeCacheInfo(keys: CacheVerifierKeysDictionary) {
+        guard let data = try? JSONEncoder().encode(keys) else {
```
corycaywood, line 133: Seems like this method is a duplicate of `writePermissionInfo`. Could you use a generic write method instead? For example:
```
private func write<T:Encodable>(keys: [String: T])
```

MiniApp/Classes/core/Storage/MiniAppKeyChain.swift
```diff
@@ -74,11 +76,78 @@ import Foundation
...
+        guard let keys = ResponseDecoder.decode(decodeType: KeysDictionary.self, data: storedData) else {
+            return nil
+        }
+
+        return keys
+    }
+
+    // MARK: - Cache Verifier Methods
+    internal func setCacheInfo(key: String, for keyId: String) {
+        var keysDic = getAllCacheKeys()
```
corycaywood, line 100: Could you add unit tests for this functionality? The RSDKUtils repo has unit tests, so maybe you can copy some of those?

MiniApp/Classes/core/Storage/MiniAppKeyChain.swift
```diff
@@ -74,11 +76,78 @@ import Foundation
...
 
-    private func write(keys: KeysDictionary) {
+    private func writePermissionInfo(keys: KeysDictionary) {
         guard let data = try? JSONEncoder().encode(keys) else {
             return
         }
+        setValueInKeyChain(data: data)
+    }
+
+    private func retrieveAllPermissions() -> KeysDictionary? {
```
corycaywood, line 86: It seems like this `MiniAppKeyChain` has too many purposes now. It's handling interactions with the KeyChain, parsing of permissions, and parsing of verification data. I think it should have just a single purpose: interacting with the KeyChain.

So maybe we should have three separate classes: `MiniAppKeyChain`, `MiniAppPermissionsStorage`, and `MiniAppVerificationStorage`. But probably that could be changed in another PR instead of this one.

What do you think?

MiniApp/Classes/core/Storage/MiniAppKeyChain.swift
```diff
@@ -74,11 +76,78 @@ import Foundation
...
+
+        guard let keys = ResponseDecoder.decode(decodeType: KeysDictionary.self, data: storedData) else {
+            return nil
+        }
+
+        return keys
+    }
+
+    // MARK: - Cache Verifier Methods
+    internal func setCacheInfo(key: String, for keyId: String) {
```
corycaywood, line 99: I think the naming of these methods isn't very clear. `Cache` could refer to many different things that we're caching in the SDK - we're also caching permissions in this same class. Maybe we could rename it with `CacheVerification`? 
- `setCacheVerificationInfo`
- `getCacheVerificationInfo`
- etc...

corycaywood (commented):
Nice, thanks for removing this library. I left a few comments below.

MiniApp/Classes/core/Storage/MiniAppKeyChain.swift
ts-leojoseph-r: Done

MiniApp/Classes/core/Storage/MiniAppKeyChain.swift
ts-leojoseph-r: That's a good idea. Done 👍

MiniApp/Classes/core/Storage/MiniAppKeyChain.swift
ts-leojoseph-r: Working..

MiniApp/Classes/core/Storage/MiniAppKeyChain.swift
ts-leojoseph-r: Done

MiniApp/Classes/core/Storage/MiniAppCacheVerifier.swift
```diff
@@ -1,27 +1,26 @@
...
         }
         let cachedFilesHash = calculateHash(appId: appId, version: version)
-        let storedHash = keystore.key(for: generateKeyId(for: appId, version: version)) ?? ""
+        let storedHash = miniAppVerificationStore.getCacheVerificationInfo(for: generateKeyId(for: appId, version: version)) ?? ""
         return cachedFilesHash == storedHash
     }
 
     func storeHash(for appId: String, version: String) {
-        keystore.removeKey(for: appId) // v2.5 legacy cleaning
+        miniAppVerificationStore.removeCacheInfo(for: appId) // v2.5 legacy cleaning
```
Climbatize, line 20: I think that is something we can get rid of with next major release, as our only known customer so far is Link and they already use v2.7.

cc @corycaywood

Climbatize (approved):
A nice refactor, LGTM

corycaywood (approved):
Looks good! 👍

PR #227, opened 2021-03-04
Feat/required optional permissions

ts-leojoseph-r:
# Description
* Added new interface `getDownloadedManifest` to return the cached MiniAppManifest
* Added new key to store MiniAppManifest
* Updated SDK to verify if the User has agreed to all required permissions before launching the mini-app
* Update MiniApp.setCustomPermissions to set only the list of custom permissions that are mentioned in the manifest.json
* Update MiniApp.getCustomPermissions to return only the list of custom permissions that are mentioned in the manifest.json

## Links
[MINI-2508](https://jira.rakuten-it.com/jira/browse/MINI-2508)

# Checklist

### DIFF
--- a/Example/Controllers/Extensions/ViewController+MiniApp.swift
+++ b/Example/Controllers/Extensions/ViewController+MiniApp.swift
@@ -66,7 +66,11 @@ extension ViewController: MiniAppNavigationDelegate {
                     self.performSegue(withIdentifier: "DisplayMiniApp", sender: nil)
                 }
             case .failure(let error):
-                self.checkSDKErrorAndDisplay(error: error)
+                self.displayAlert(title: NSLocalizedString("error_title", comment: ""),
+                                  message: NSLocalizedString("error_miniapp_download_message", comment: "") + ".\n\(error.localizedDescription)",
+                                  dismissController: true) { _ in
+                    self.fetchAppList(inBackground: true)
+                }
                 print("Errored: ", error.localizedDescription)
             }
         }, messageInterface: self, adsDisplayer: adsDisplayer)
@@ -92,19 +96,6 @@ extension ViewController: MiniAppNavigationDelegate {
             } else {
                 self.fetchMiniAppUsingId(title: NSLocalizedString("error_invalid_miniapp_id", comment: ""), message: NSLocalizedString("input_valid_miniapp_title", comment: ""))
             }
-        }
-    }
-
-    func checkSDKErrorAndDisplay(error: MASDKError) {
-        var errorMessage: String = ""
-        switch error {
-        case .metaDataFailure:
-            errorMessage = NSLocalizedString("error_miniapp_download_message", comment: "") + ".\n\(error.localizedDescription)"
-        default:
-            errorMessage = NSLocalizedString("error_miniapp_download_message", comment: "")
-        }
-        self.displayAlert(title: NSLocalizedString("error_title", comment: ""), message: errorMessage, dismissController: true) { _ in
-            self.fetchAppList(inBackground: true)
         }
     }
 }
--- a/Example/Controllers/Extensions/ViewController+MiniApp.swift
+++ b/Example/Controllers/Extensions/ViewController+MiniApp.swift
@@ -66,9 +66,7 @@ extension ViewController: MiniAppNavigationDelegate {
                     self.performSegue(withIdentifier: "DisplayMiniApp", sender: nil)
                 }
             case .failure(let error):
-                self.displayAlert(title: NSLocalizedString("error_title", comment: ""), message: NSLocalizedString("error_miniapp_download_message", comment: ""), dismissController: true) { _ in
-                    self.fetchAppList(inBackground: true)
-                }
+                self.checkSDKErrorAndDisplay(error: error)
                 print("Errored: ", error.localizedDescription)
             }
         }, messageInterface: self, adsDisplayer: adsDisplayer)
@@ -94,6 +92,19 @@ extension ViewController: MiniAppNavigationDelegate {
             } else {
                 self.fetchMiniAppUsingId(title: NSLocalizedString("error_invalid_miniapp_id", comment: ""), message: NSLocalizedString("input_valid_miniapp_title", comment: ""))
             }
+        }
+    }
+
+    func checkSDKErrorAndDisplay(error: MASDKError) {
+        var errorMessage: String = ""
+        switch error {
+        case .metaDataFailure:
+            errorMessage = NSLocalizedString("error_miniapp_download_message", comment: "") + ". \nPlease make sure user agreed to all required permissions from Meta-data"
+        default:
+            errorMessage = NSLocalizedString("error_miniapp_download_message", comment: "")
+        }
+        self.displayAlert(title: NSLocalizedString("error_title", comment: ""), message: errorMessage, dismissController: true) { _ in
+            self.fetchAppList(inBackground: true)
         }
     }
 }
--- a/Example/Controllers/Extensions/ViewController+MiniApp.swift
+++ b/Example/Controllers/Extensions/ViewController+MiniApp.swift
@@ -99,7 +99,7 @@ extension ViewController: MiniAppNavigationDelegate {
         var errorMessage: String = ""
         switch error {
         case .metaDataFailure:
-            errorMessage = NSLocalizedString("error_miniapp_download_message", comment: "") + ". \nPlease make sure user agreed to all required permissions from Meta-data"
+            errorMessage = NSLocalizedString("error_miniapp_download_message", comment: "") + ".\n\(error.localizedDescription)"
         default:
             errorMessage = NSLocalizedString("error_miniapp_download_message", comment: "")
         }
--- a/MiniApp/Classes/core/MiniApp.swift
+++ b/MiniApp/Classes/core/MiniApp.swift
@@ -107,6 +107,14 @@ public class MiniApp: NSObject {
     public func getMiniAppManifest(miniAppId: String, miniAppVersion: String, completionHandler: @escaping (Result<MiniAppManifest, MASDKError>) -> Void) {
         return realMiniApp.retrieveMiniAppMetaData(appId: miniAppId, version: miniAppVersion, completionHandler: completionHandler)
     }
+
+    /// Method to return the cached meta-data information of a mini-app from the Keychain
+    /// - Parameters:
+    ///   - miniAppId:  Mini AppId String value
+    /// - Returns: MiniAppManifest object info from the cache
+    public func getDownloadedManifest(miniAppId: String) -> MiniAppManifest? {
+        return realMiniApp.getCachedManifestData(appId: miniAppId)
+    }
 }
 
 // MARK: - Testing
--- a/MiniApp/Classes/core/MiniApp.swift
+++ b/MiniApp/Classes/core/MiniApp.swift
@@ -108,10 +108,10 @@ public class MiniApp: NSObject {
         return realMiniApp.retrieveMiniAppMetaData(appId: miniAppId, version: miniAppVersion, completionHandler: completionHandler)
     }
 
-    /// Method to return the cached meta-data information of a mini-app from the Keychain
+    /// Method to return the cached meta-data information of a mini-app
     /// - Parameters:
     ///   - miniAppId:  Mini AppId String value
-    /// - Returns: MiniAppManifest object info from the cache
+    /// - Returns: MiniAppManifest object info from the cache, Returns nil, if the mini-app is not downloaded already.
     public func getDownloadedManifest(miniAppId: String) -> MiniAppManifest? {
         return realMiniApp.getCachedManifestData(appId: miniAppId)
     }
--- a/MiniApp/Classes/core/RealMiniApp.swift
+++ b/MiniApp/Classes/core/RealMiniApp.swift
@@ -7,7 +7,6 @@ internal class RealMiniApp {
     var miniAppStatus: MiniAppStatus
     var miniAppPermissionStorage: MiniAppPermissionsStorage
     var miniAppManifestStorage: MAManifestStorage
-    var miniAppManifest: MiniAppManifest?
     let offlineErrorCodeList: [Int] = [NSURLErrorNotConnectedToInternet, NSURLErrorTimedOut]
 
     convenience init() {
@@ -136,6 +135,8 @@ internal class RealMiniApp {
                         completionHandler: @escaping (Result<MiniAppDisplayDelegate, Error>) -> Void,
                         messageInterface: MiniAppMessageDelegate? = nil,
                         adsDisplayer: MiniAppAdDisplayer? = nil) {
+        self.miniAppStatus.setDownloadStatus(true, appId: appInfo.id, versionId: appInfo.version.versionId)
+        self.miniAppStatus.setCachedVersion(appInfo.version.versionId, for: appInfo.id)
         isRequiredPermissionsAllowed(
             appId: appInfo.id,
             versionId: appInfo.version.versionId) { (result) in
@@ -149,8 +150,6 @@ internal class RealMiniApp {
                                                                                    queryParams: queryParams,
                                                                                    hostAppMessageDelegate: messageInterface ?? self,
                                                                                    adsDisplayer: adsDisplayer)
-                        self.miniAppStatus.setDownloadStatus(true, appId: appInfo.id, versionId: appInfo.version.versionId)
-                        self.miniAppStatus.setCachedVersion(appInfo.version.versionId, for: appInfo.id)
                         completionHandler(.success(miniAppDisplayProtocol))
                     }
             case .failure(let error):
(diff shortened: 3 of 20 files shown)


MiniApp/Classes/core/RealMiniApp.swift
```diff
@@ -212,6 +227,54 @@ internal class RealMiniApp {
...
         self.miniAppStatus.removeUnusedCustomPermissions()
     }
+
+    /// Method to check if all the required permissions mentioned in the manifest.json is agreed by the user.
+    /// - Parameters:
+    ///   - appId: MiniApp ID
+    ///   - versionId: Specific VersionID of a MiniApp
+    ///   - completionHandler: Handler that returns whether user agreed to required permissions or not.
+    func isRequiredPermissionsAllowed(appId: String, versionId: String, completionHandler: @escaping (Result<Bool, MASDKError>) -> Void) {
+        retrieveMiniAppMetaData(appId: appId, version: versionId) { (result) in
```
corycaywood, line 237: The Host App should have already retrieved the manifest before calling `create`, so this means we will end up making two API requests for the exact same thing?

Maybe we could make a class for in-memory caching the manifest from the API? For example, when the Host App calls `getMiniAppManifest`, we will cache the manifest in-memory (not to a file). And then when they call `create`, we can check the in-memory cache whether we already downloaded the manifest for that version. If not, then we can call the API to download it. The Host App should typically only be calling `getMiniAppManifest` just before they call `create`, so we would probably only need to cache the latest manifest response from `getMiniAppManifest`.

What do you think @leorjoseph @khairul-alam-licon ?

MiniApp/Classes/core/RealMiniApp.swift
```diff
@@ -212,6 +227,54 @@ internal class RealMiniApp {
     func cleanUpKeychain() {
         self.miniAppStatus.removeUnusedCustomPermissions()
     }
+
+    /// Method to check if all the required permissions mentioned in the manifest.json is agreed by the user.
+    /// - Parameters:
+    ///   - appId: MiniApp ID
+    ///   - versionId: Specific VersionID of a MiniApp
+    ///   - completionHandler: Handler that returns whether user agreed to required permissions or not.
+    func isRequiredPermissionsAllowed(appId: String, versionId: String, completionHandler: @escaping (Result<Bool, MASDKError>) -> Void) {
```
corycaywood, line 236: The `RealMiniApp` class is intended to just be the place where the various components are connected. But it feels like it's turning into a "god class" which has too much functionality. Is there any way you could refactor some of this permissions functionality into another class? 

Sorry, I don't really have a good suggestion on this. But basically, I feel like `RealMiniApp` is getting too large. Maybe it's something we can add as "technical debt" and address in another PR?

MiniApp/Classes/core/RealMiniApp.swift
```diff
@@ -128,18 +130,31 @@ internal class RealMiniApp {
...
-                                                                       adsDisplayer: adsDisplayer)
-            self.miniAppStatus.setDownloadStatus(true, appId: appInfo.id, versionId: appInfo.version.versionId)
-            self.miniAppStatus.setCachedVersion(appInfo.version.versionId, for: appInfo.id)
-            completionHandler(.success(miniAppDisplayProtocol))
+    func getMiniAppView(appInfo: MiniAppInfo,
+                        queryParams: String? = nil,
+                        completionHandler: @escaping (Result<MiniAppDisplayDelegate, Error>) -> Void,
+                        messageInterface: MiniAppMessageDelegate? = nil,
+                        adsDisplayer: MiniAppAdDisplayer? = nil) {
+        isRequiredPermissionsAllowed(
```
corycaywood, line 138: This call to `isRequiredPermissionsAllowed` is buried inside several other function calls, so it makes it more difficult to understand the behavior when creating a Mini App. It seems like the code would be easier to read if this call was moved higher, i.e. into `createMiniApp`. But I'm not sure if you can keep the exact same behavior if you move it somewhere else?

corycaywood (commented):
Looking good! I left a few suggestions below.

MiniApp/Classes/core/RealMiniApp.swift
```diff
@@ -212,6 +227,54 @@ internal class RealMiniApp {
...
+            case .success(let manifest):
+                guard let requiredPermissions = manifest.requiredPermissions, requiredPermissions.count > 0 else {
+                    completionHandler(.success(true))
+                    return
+                }
+                self.verifyRequiredPermissions(appId: appId,
+                                               requiredPermissions: requiredPermissions,
+                                               completionHandler: completionHandler)
+            case .failure(let error):
+                completionHandler(.failure(.fromError(error: error)))
```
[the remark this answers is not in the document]
ts-leojoseph-r: Yeah, that'r right. But in this scenario, `getMiniAppView` has to return Error, whereas methods called from `getMiniAppView` and also methods which  call `getMiniAppView` needs MASDKErrror. I am trying to overcome this issue now. Kinda complicated.

MiniApp/Classes/core/RealMiniApp.swift
```diff
@@ -128,18 +130,31 @@ internal class RealMiniApp {
...
+                                                                                   miniAppTitle: appInfo.displayName ?? "Mini app",
+                                                                                   queryParams: queryParams,
+                                                                                   hostAppMessageDelegate: messageInterface ?? self,
+                                                                                   adsDisplayer: adsDisplayer)
+                        self.miniAppStatus.setDownloadStatus(true, appId: appInfo.id, versionId: appInfo.version.versionId)
+                        self.miniAppStatus.setCachedVersion(appInfo.version.versionId, for: appInfo.id)
+                        completionHandler(.success(miniAppDisplayProtocol))
+                    }
+            case .failure:
+                completionHandler(.failure(MASDKError.metaDataFailure))
```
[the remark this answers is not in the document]
ts-leojoseph-r: Once the above Error issue resolved, I will add the same.

MiniApp/Classes/core/RealMiniApp.swift
ts-leojoseph-r: Yeah, I felt the same. We can add extensions or move few functions into a separate class. But don't want to do this close to the release. Will work on this after the release

MiniApp/Classes/core/RealMiniApp.swift
khairul-alam-licon: Yeah, in Android side, I was caching the latest manifest which is fetched by `getMiniAppManifest` from api. 
@leorjoseph Do you think, it's sound good?

ts-leojoseph-r: It's done already @khairul-alam-licon. Also, @leorjoseph is not my official ID. Please use @ts-leojoseph-r

MiniApp/Classes/core/RealMiniApp.swift
Climbatize: I think using an extension would not be different from having this big class. We should move all the methods that don't need to be instanced to class methods in a dedicated tools class.

MiniApp/Classes/core/RealMiniApp.swift
Climbatize: come on @leorjoseph !

MiniApp/Classes/core/RealMiniApp.swift
```diff
@@ -205,13 +240,69 @@ internal class RealMiniApp {
...
+    ///   - requiredPermissions: List of required Custom permissions that is defined by the Mini App
+    ///   - completionHandler: Handler that returns whether user agreed to required permissions or not.
+    func verifyRequiredPermissions(appId: String,
+                                   requiredPermissions: [MASDKCustomPermissionModel],
+                                   completionHandler: @escaping (Result<Bool, MASDKError>) -> Void) {
+        let storedCustomPermissions = self.miniAppPermissionStorage.getCustomPermissions(forMiniApp: appId)
+        let filtered = storedCustomPermissions.filter {
+            requiredPermissions.contains($0)
+        }
+        if filtered.count == requiredPermissions.count && filtered.allSatisfy({ $0.isPermissionGranted.boolValue == true }) {
```
Climbatize, line 293: I thing you could add the `allSatisfy` _(awesome language feature!)_ condition to the filter above

MiniApp/Classes/core/RealMiniApp.swift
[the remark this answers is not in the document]
ts-leojoseph-r: Updated

MiniApp/Classes/core/RealMiniApp.swift
ts-leojoseph-r: Yes, we can. But I wanted to avoid any refactoring now, especially when we are this close to release.

MiniApp/Classes/core/RealMiniApp.swift
ts-leojoseph-r: But I would suggest that we call this `isRequiredPermissionsAllowed` method right even before we download.

MiniApp/Classes/core/RealMiniApp.swift
ts-leojoseph-r: No, we shouldn't. Stored Permissions contains both required and optional permissions So we need to filter only required permissions and make sure only they have all permissions `.allowed`

corycaywood (commented):
I left a few more minor comments below. But please see the comment about `getValidManifest`. I'm not sure how to best handle this.

MiniApp/Classes/core/RealMiniApp.swift
corycaywood: I think that `isRequiredPermissionsAllowed` is unrelated to downloading, so it doesn't make sense to call it before download. For example, even if the mini app is already downloaded, we still need to check `isRequiredPermissionsAllowed` because the user could have revoked permissions in the permission settings screen.

Anyways, I think it's okay to leave this here - it might need a lot of refactoring to move it somewhere else.

Example/Controllers/Extensions/ViewController+MiniApp.swift
```diff
@@ -96,4 +94,17 @@ extension ViewController: MiniAppNavigationDelegate {
             }
         }
     }
+
+    func checkSDKErrorAndDisplay(error: MASDKError) {
+        var errorMessage: String = ""
+        switch error {
+        case .metaDataFailure:
```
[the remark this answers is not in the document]
ts-leojoseph-r: Ah, yes. This block also returns the error if there is an issue with Download, so I guess it's better to handle them. 

Also, I was thinking, shouldn't we display the first-time UI again if there is an error with the meta-data failure? Which is also the requirement for https://jira.rakuten-it.com/jira/browse/MINI-2509 right ?

MiniApp/Classes/core/MiniApp.swift
```diff
@@ -107,6 +107,14 @@ public class MiniApp: NSObject {
     public func getMiniAppManifest(miniAppId: String, miniAppVersion: String, completionHandler: @escaping (Result<MiniAppManifest, MASDKError>) -> Void) {
         return realMiniApp.retrieveMiniAppMetaData(appId: miniAppId, version: miniAppVersion, completionHandler: completionHandler)
     }
+
+    /// Method to return the cached meta-data information of a mini-app from the Keychain
+    /// - Parameters:
+    ///   - miniAppId:  Mini AppId String value
+    /// - Returns: MiniAppManifest object info from the cache
```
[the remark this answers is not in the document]
ts-leojoseph-r: Done

MiniApp/Classes/core/RealMiniApp.swift
```diff
@@ -173,11 +189,30 @@ internal class RealMiniApp {
...
+        }
+        let manifestCustomPermissions = (manifestData.requiredPermissions ?? []) + (manifestData.optionalPermissions ?? [])
+        let filtered = cachedPermissions.filter {
+            manifestCustomPermissions.contains($0)
+        }
+        return filtered
+    }
+
+    func getValidManifest(forMiniApp id: String) -> MiniAppManifest? {
+        guard let manifest = self.miniAppManifest else {
```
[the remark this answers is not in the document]
ts-leojoseph-r: Done

MiniApp/Classes/core/RealMiniApp.swift
```diff
@@ -205,13 +232,78 @@ internal class RealMiniApp {
...
+                                   miniAppManifest: MiniAppManifest?,
+                                   completionHandler: @escaping (Result<Bool, MASDKError>) -> Void) {
+        guard let manifestData = miniAppManifest, let requiredPermissions = manifestData.requiredPermissions else {
+            return completionHandler(.success(true))
+        }
+        let storedCustomPermissions = self.miniAppPermissionStorage.getCustomPermissions(forMiniApp: appId)
+        let filtered = storedCustomPermissions.filter {
+            requiredPermissions.contains($0)
+        }
+        if filtered.count == requiredPermissions.count && filtered.allSatisfy({ $0.isPermissionGranted.boolValue == true }) {
```
corycaywood, line 293: I think we also need to update ` MiniAppPermissionStorage` somewhere in this function - i.e. filter the stored custom permissions so it only has the entries from `manifest.json`. 

I think I had mentioned doing this after verifying the required permissions, but maybe it makes more sense to do this just before this `if` check?

MiniApp/Classes/core/RealMiniApp.swift
```diff
@@ -205,13 +232,78 @@ internal class RealMiniApp {
...
+            self.miniAppManifestStorage.saveManifestInfo(forMiniApp: appId, manifest: CachedMetaData(version: versionId, miniAppManifest: manifestData))
```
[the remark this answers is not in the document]
ts-leojoseph-r: Makes more sense 👍

MiniApp/Classes/core/Storage/MAManifestStorage.swift
```diff
@@ -0,0 +1,59 @@
+import Foundation
+import CommonCrypto
+
+internal class MAManifestStorage {
+
+    typealias KeysDictionary = [String: CachedMetaData]
+    let keychainStore = MiniAppKeyChain(serviceName: .customPermission)
```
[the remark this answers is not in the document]
ts-leojoseph-r: Ahh, savior 👍

MiniApp/Classes/core/RealMiniApp.swift
ts-leojoseph-r: Ok, I made use of the filterCustomPermissions method which filters and the same is stored now

corycaywood (commented):
Looks good, but I just noticed one issue:

Do a clean launch of the Demo App then launch an old version of the JS Sample App which doesn't have a manifest.json.

Now if you check the Permissions settings screen, the mini app was granted all permissions automatically, but it should have been granted no permissions because it doesn't have a manifest. 

Not sure if maybe this is an issue with the demo app calling `MiniApp.storeCustomPermissions` when it shouldn't?

corycaywood (approved):
LGTM 👍

Climbatize (approved):

USERGUIDE.md
```diff
@@ -75,6 +75,7 @@ Config.userDefaults?.set("MY_CUSTOM_ID", forKey: Config.Key.subscriptionKey.rawV
 * [Load the Mini App list](#load-miniapp-list)
 * [Get a MiniAppInfo](#get-mini-appinfo)
 * [Getting a MiniApp meta-data](#get-mini-meta-data)
+* [How to use MiniApp meta-data](#how-to-use-meta-data)
```
Climbatize, line 78: there should be a MiniApp meta-data section with these 2 subsections

Tests/Unit/RealMiniAppTests.swift
```diff
@@ -145,7 +150,7 @@ class RealMiniAppTests: QuickSpec {
                 it("will return valid Mini App View instance with a default hostAppMessageDelegate and getUniqueId() will return an error message") {
                     let responseString = """
                     [{
-                        "id": "123",
+                        "id": "app-id-test",
```
Climbatize, line 153: can you refactor it with a class or global variable?

Climbatize (approved):
LGTM. Some trivial comments, treat them if you feel like doing it

USERGUIDE.md
```diff
@@ -484,6 +486,19 @@ MiniApp.shared().getMiniAppManifest(miniAppId: miniAppId, miniAppVersion: miniAp
...
+### How to use `MiniApp meta-data` :
+---
+
+SDK internally checks if the User has responded to the list of required/optional permissions that is enforced by the mini-app. So host app SHOULD make sure that the user is prompted with necessary custom permissions and get the response back from the user.
+
+Before calling the MiniApp.create, host app should make sure the following things are done,
+
+* Retrieve the meta-data for the mini-app using [getMiniAppManifest](#get-mini-meta-data)
+* Display/Prompt the list of required & optional permissions to the user and the user response should be stored using [MiniApp.setCustomPermissions](#store-custom-permission)
+* Call [MiniApp.create](#create-mini-app) to start downloading the mini-app
```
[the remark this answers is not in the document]
ts-leojoseph-r: Done

USERGUIDE.md
```diff
@@ -484,6 +486,19 @@ MiniApp.shared().getMiniAppManifest(miniAppId: miniAppId, miniAppVersion: miniAp
...
 ```
 
+<a id="get-mini-meta-data"></a>
+
+### How to use `MiniApp meta-data` :
+---
+
+SDK internally checks if the User has responded to the list of required/optional permissions that is enforced by the mini-app. So host app SHOULD make sure that the user is prompted with necessary custom permissions and get the response back from the user.
+
+Before calling the MiniApp.create, host app should make sure the following things are done,
```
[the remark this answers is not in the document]
ts-leojoseph-r: Done

USERGUIDE.md
```diff
@@ -484,6 +486,19 @@ MiniApp.shared().getMiniAppManifest(miniAppId: miniAppId, miniAppVersion: miniAp
...
+
+* Retrieve the meta-data for the mini-app using [getMiniAppManifest](#get-mini-meta-data)
```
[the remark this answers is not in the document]
ts-leojoseph-r: Updated Mini App text, but I think [getMiniAppManifest] is better than `getMiniAppManifest` as the first will show more like the Hyperlink

USERGUIDE.md
ts-leojoseph-r: Done

Tests/Unit/RealMiniAppTests.swift
ts-leojoseph-r: Done

PR #228, opened 2021-03-05
Access Token Audience/Scopes

Climbatize:
# Description
Implementation of access token scopes in manifest.json

## Links
MINI-2252

# Checklist


corycaywood (commented):
I left a few comments below. Sorry, I think you did more work than you needed to 😣  The Host App doesn't need to be able to set `AccessTokenScopes` entries. The scopes are always set by the mini app in their manifest.json. So this means we don't need to store the scopes separately, and we can just access them from the manifest.

MiniApp/Classes/core/JavascriptBridge/MiniAppScriptMessageHandler.swift
```diff
@@ -339,8 +335,19 @@ internal class MiniAppScriptMessageHandler: NSObject, WKScriptMessageHandler {
...
-        hostAppMessageDelegate?.getAccessToken(miniAppId: self.miniAppId) { (result) in
+    func fetchTokenDetails(callbackId: String, for requestParam: RequestParameters?) {
+        var accessTokenPermission: MASDKAccessTokenPermission?
+        if let audience = requestParam?.audience {
+            accessTokenPermission = MASDKAccessTokenPermission(audience: audience, scopes: [])
+            guard let scopes = requestParam?.scopes else {
+                return sendScopeError(callbackId: callbackId, type: .scopeError)
+            }
+            accessTokenPermission?.scopes = scopes
+            if !(miniAppScope?.contains(accessTokenPermission!) ?? false) {
```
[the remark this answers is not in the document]
Climbatize: Oh for that it's exactly what it's checking, I've overriden the equality operator of MASDKAccessTokenPermission to check subset, but I realize now it's not very clever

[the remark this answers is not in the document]
corycaywood: Ah, I see! Sorry, I didn't notice that. Hmm, yeah having it in the equality operator seems confusing - equality should mean that the `audience` and `scopes` match exactly. And it seems like we might want to use actual equality for something else.

Maybe you could add a separate function to `MASDKAccessTokenPermission` for this? Something like `containsAll` or `containsSubSet`?

MiniApp/Classes/core/MiniApp.swift
```diff
@@ -67,6 +74,14 @@ public class MiniApp: NSObject {
         return realMiniApp.storeCustomPermissions(forMiniApp: appId, permissionList: permissionList)
     }
 
+    /// Cache the token scopes for a given MiniApp ID
+    /// - Parameters:
+    ///   - appId: Mini AppId String value
+    ///   - scopeList: List of MASDKAccessTokenPermission struct that contains the audience and scopes
+    public func setScopes(forMiniApp appId: String, scopeList: [MASDKAccessTokenPermission]) {
```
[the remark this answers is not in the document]
Climbatize: removed

MiniApp/Classes/core/MiniApp.swift
```diff
@@ -77,6 +92,16 @@ public class MiniApp: NSObject {
         return []
     }
 
+    /// Get the list of supported scopes for a given Mini app ID
+    /// - Parameter appId: Mini AppId String value
+    /// - Returns: List of MASDKAccessTokenPermission that contains the details of every scopes.
+    public func getScopes(forMiniApp appId: String) -> [MASDKAccessTokenPermission] {
```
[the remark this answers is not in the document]
Climbatize: removed

MiniApp/Classes/core/Storage/MiniAppScopeStorage.swift
```diff
@@ -0,0 +1,62 @@
+import Foundation
+
+internal class MiniAppScopeStorage {
```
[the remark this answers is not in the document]
Climbatize: no need indeed, removed

MiniApp/Classes/core/JavascriptBridge/MiniAppScriptMessageHandler.swift
```diff
@@ -18,13 +17,15 @@ internal class MiniAppScriptMessageHandler: NSObject, WKScriptMessageHandler {
     var userAlreadyRespondedRequestList = [MASDKCustomPermissionModel]()
     var cachedUnknownCustomPermissionRequest = [MiniAppCustomPermissionsListResponse]()
     var miniAppKeyStore = MiniAppPermissionsStorage()
+    var miniAppScope: [MASDKAccessTokenPermission]? = [MASDKAccessTokenPermission]()
```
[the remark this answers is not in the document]
Climbatize: replaced by MiniAppManifest

MiniApp/Classes/core/JavascriptBridge/MiniAppScriptMessageHandler.swift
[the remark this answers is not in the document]
Climbatize: I fixed the equality operator and created a dedicated method in `MASDKAccessTokenPermission`

MiniApp/Classes/core/JavascriptBridge/MiniAppScriptMessageHandler.swift
```diff
@@ -340,8 +335,20 @@ internal class MiniAppScriptMessageHandler: NSObject, WKScriptMessageHandler {
...
+        if let audience = requestParam?.audience { // if we request a token for a specific audience
+            accessTokenPermission = MASDKAccessTokenPermission(audience: audience, scopes: [])
+            if let scopes = requestParam?.scopes { // we need a specific set of scopes
+                accessTokenPermission?.scopes = scopes
+            } else { return sendScopeError(callbackId: callbackId, type: .scopeError) }
+            if let accessTokenPermissions = miniAppManifest?.accessTokenPermissions { // we check that the Mini App manages scopes
+                if !(accessTokenPermission!.isPartOf(accessTokenPermissions)) { // and that these scopes are included in the Manifest
+                    return sendScopeError(callbackId: callbackId, type: .audienceError)
+                }
+            } else { return sendScopeError(callbackId: callbackId, type: .audienceError) }
```
ts-leojoseph-r, line 349: You are throwing the same error, is it possible to make them as one?

MiniApp/Classes/core/RealMiniApp.swift
```diff
@@ -106,59 +124,58 @@ internal class RealMiniApp {
...
+                                                                       versionId: appInfo.version.versionId,
+                                                                       projectId: self.miniAppClient.environment.projectId,
+                                                                       miniAppTitle: appInfo.displayName ?? "Mini app",
+                                                                       manifest: manifest,
+                                                                       queryParams: queryParams,
+                                                                       hostAppMessageDelegate: messageInterface ?? self,
+                                                                       adsDisplayer: adsDisplayer)
+            self.miniAppStatus.setDownloadStatus(true, appId: appInfo.id, versionId: appInfo.version.versionId)
+            self.miniAppStatus.setCachedVersion(appInfo.version.versionId, for: appInfo.id)
+            completionHandler(.success(miniAppDisplayProtocol))
```
ts-leojoseph-r, line 173: isRequiredPermissionsAllowed() is removed here. We need to verify if all required permissions are granted before returning the view.

MiniApp/Classes/core/Models/MiniAppManifest.swift
```diff
@@ -31,6 +32,32 @@ internal struct MACustomPermissionsResponse: Decodable {
     }
 }
 
+public struct MASDKAccessTokenPermission: Codable, Equatable, Hashable {
```
ts-leojoseph-r, line 35: `MASDKAccessTokenPermission` -> I guess this could create some confusion, what do you think? May be `MASDKAccessTokenScopes` / `MASDKAccessTokenMetaData` / `MASDKAccessTokenModel` ?

MiniApp/Classes/core/Models/MiniAppManifest.swift
```diff
@@ -31,6 +32,32 @@ internal struct MACustomPermissionsResponse: Decodable {
...
+
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(audience)
+    }
+
+    public static func == (lhs: MASDKAccessTokenPermission, rhs: MASDKAccessTokenPermission) -> Bool {
+        lhs.audience == rhs.audience && rhs.scopes.sorted() == lhs.scopes.sorted()
+    }
+
+    public func isPartOf(_ fullScopes: [MASDKAccessTokenPermission]) -> Bool {
```
ts-leojoseph-r, line 52: Some documentation for public interfaces would be great to have

ts-leojoseph-r (changes requested):
Overall looks good. I have added few comments, please check.

Example/Controllers/MAFirstLaunchController.swift
```diff
@@ -59,7 +59,8 @@ class MAFirstLaunchController: UIViewController {
     }
 
     @IBAction func acceptButtonPressed(_ sender: UIButton) {
-        MiniApp.shared().setCustomPermissions(forMiniApp: miniAppInfo?.id ?? "", permissionList: permissionsCollections ?? [])
+        let appId = miniAppInfo?.id ?? ""
+        MiniApp.shared().setCustomPermissions(forMiniApp: appId, permissionList: permissionsCollections ?? [])
```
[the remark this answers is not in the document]
Climbatize: This my friend is what they call a bad merge 😭

Example/Models/UserProfileModel.swift
```diff
@@ -81,7 +81,7 @@ func saveTokenInfo(accessToken: String, expiryDate: Date, forKey key: String = "
         return false
 }
 
-func getTokenInfo(key: String = "AccessTokenInfo") -> AccessTokenInfo? {
+func getTokenInfo(key: String = "AccessTokenInfo", miniApp: String? = nil, audience: String? = nil, scope: [String]? = nil) -> AccessTokenInfo? {
```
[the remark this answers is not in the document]
Climbatize: Again a remaining of my previous version of this PR 😭

MiniApp/Classes/core/JavascriptBridge/MiniAppScriptMessageHandler.swift
Climbatize: yeah, maybe if I think a little more about it

MiniApp/Classes/core/Models/MiniAppManifest.swift
Climbatize: I have a preference for the Scopes one, but the MetaData entry is called `AccessTokenPermission`. What do you think @corycaywood ?

MiniApp/Classes/core/Models/MiniAppManifest.swift
Climbatize: of course. I just thought the name was self explanatory ^^

MiniApp/Classes/core/RealMiniApp.swift
Climbatize: sorry, I don't know what happened, a bad merge? please help me to solve this case 😔

corycaywood: @Climbatize The Access Token permission changes were merged to the `feature` branch. So actually I think you should have been working on top of that branch instead of `master`. 

But now that we've already merged the meta-data feature to `candidate`, I think we can work on `master` for the access token feature. So probably first we need to merge `feature` to `master` and then you can rebase your branch on `master`.

MiniApp/Classes/core/Models/MiniAppManifest.swift
corycaywood: Maybe it's just missing an "s": `MASDKAccessTokenPermissions`?

The key name in the manifest is `accessTokenPermissions`, so it seems a little confusing if the name doesn't match.

corycaywood: Oh, sorry I just noticed this is also being used in `MiniAppUserInfoDelegate.getAccessToken`. Yeah, I think it makes more sense to change it to `MASDKAccessTokenScopes` because it's being used by the Mini App to specify which scopes they want.

However, in the case that it's used as an array `[MASDKAccessTokenPermission]`  (such as above in this class), it should keep the name `accessTokenPermissions`.

MiniApp/Classes/core/Models/MiniAppManifest.swift
corycaywood: Could you also add documentation for the class `MASDKAccessTokenScopes` too?

USERGUIDE.md
```diff
@@ -481,15 +497,34 @@ Host app will use the defined interfaces to retrieve these details from manifest
 Retrieve the meta-data of a MiniApp using the following method,
 
 ```swift
+var theManifest: MiniAppManifest?
+
 MiniApp.shared().getMiniAppManifest(miniAppId: miniAppId, miniAppVersion: miniAppVersionId) { (result) in
     switch result {
         case .success(let manifestData):
             // Retrieve the custom key/value pair like the following.
+            theManifest = manifestData
```
corycaywood, line 506: Looks like `theManifest` isn't being used for anything in this example?

MiniApp/Classes/core/Models/MiniAppManifest.swift
Climbatize: Already planned it @corycaywood :)

USERGUIDE.md
Climbatize: Not really, I remove it

Example/Controllers/Extensions/ViewController+MiniAppUserInfoProtocol.swift
```diff
@@ -23,9 +23,11 @@ extension ViewController {
         return contactList
     }
 
-    func getAccessToken(miniAppId: String, completionHandler: @escaping (Result<MATokenInfo, MASDKCustomPermissionError>) -> Void) {
+    func getAccessToken(miniAppId: String,
+                        scopes: MASDKAccessTokenScopes?,
+                        completionHandler: @escaping (Result<MATokenInfo, MASDKCustomPermissionError>) -> Void) {
```
corycaywood, line 28: `MATokenInfo` also needs to be updated to contain the `audience` and `scopes`. This information should be passed back to the JS side so they know what scopes they have for that token (for example, the Host App might have removed some of the scopes if they were invalid).

corycaywood (commented):
Looks good, just one comment below.

Climbatize (commented):
@ts-leojoseph-r , @corycaywood , I added sample app configuration to save audience and scope into settings

MiniApp/Classes/core/RealMiniApp.swift
```diff
@@ -47,41 +47,58 @@ internal class RealMiniApp {
     /// will be passed on to Downloader class (which will check whether the mini app is downloaded already if not, it will download)
     /// - Parameters:
     ///   - appInfo: Miniapp Info object
+    ///   - manifest: optional Mini App manifest containing permissions and other metadata
     ///   - queryParams: Optional Query parameters that the host app would like to share while creating a mini app
     ///   - completionHandler: Completion Handler that needed to pass back the MiniAppDisplayProtocol
     ///   - messageInterface: Miniapp communication protocol object.
     ///   - adsDisplayer: a delegate that will handle Miniapp ads requests
-    func createMiniApp(appInfo: MiniAppInfo, queryParams: String? = nil, completionHandler: @escaping (Result<MiniAppDisplayDelegate, Error>) -> Void, messageInterface: MiniAppMessageDelegate? = nil, adsDisplayer: MiniAppAdDisplayer? = nil) {
+    func createMiniApp(appInfo: MiniAppInfo, manifest: MiniAppManifest? = nil, queryParams: String? = nil, completionHandler: @escaping (Result<MiniAppDisplayDelegate, Error>) -> Void, messageInterface: MiniAppMessageDelegate? = nil, adsDisplayer: MiniAppAdDisplayer? = nil) {
```
ts-leojoseph-r, line 55: I don't understand the need for passing the `manifest` object from the Host app. Irrespective of what manifest sent from the Host app, SDK will be checking from platform/cache right? Then we won't be needing this one.

Or my understanding is wrong here ?

MiniApp/Classes/core/RealMiniApp.swift
```diff
@@ -47,41 +47,58 @@ internal class RealMiniApp {
...
+                        queryParams: queryParams,
+                        completionHandler: completionHandler,
+                        messageInterface: messageInterface,
+                        adsDisplayer: adsDisplayer)
             } }
     }
 
     func createMiniApp(appId: String, version: String? = nil, queryParams: String? = nil, completionHandler: @escaping (Result<MiniAppDisplayDelegate, MASDKError>) -> Void, messageInterface: MiniAppMessageDelegate? = nil, adsDisplayer: MiniAppAdDisplayer? = nil) {
         getMiniApp(miniAppId: appId, miniAppVersion: version) { (result) in
+            let manifest = self.miniAppManifestStorage.getManifestInfo(forMiniApp: appId)?.miniAppManifest
```
ts-leojoseph-r, line 89: When the mini-app is downloaded for the first time, this manifest object would be nil. Also, I see you are passing this `manifest` object all the way to `func getMiniAppView()` method.

In that case, you could retrieve the manifest here -> https://github.com/rakutentech/ios-miniapp/pull/228/files#diff-af0f0dfc2aba552e5313b2588ffed2ac98cffd616f7caca2105ac7660572fc6cR167

Doing so, it will give you the latest manifest and also you can avoid passing `manifest` object

MiniApp/Classes/core/RealMiniApp.swift
```diff
@@ -106,59 +124,67 @@ internal class RealMiniApp {
...
+                    let miniAppDisplayProtocol = self.displayer.getMiniAppView(miniAppId: appInfo.id,
+                                                                               versionId: appInfo.version.versionId,
+                                                                               projectId: self.miniAppClient.environment.projectId,
+                                                                               miniAppTitle: appInfo.displayName ?? "Mini app",
+                                                                               manifest: manifest,
+                                                                               queryParams: queryParams,
+                                                                               hostAppMessageDelegate: messageInterface ?? self,
+                                                                               adsDisplayer: adsDisplayer)
+                    self.miniAppStatus.setDownloadStatus(true, appId: appInfo.id, versionId: appInfo.version.versionId)
+                    self.miniAppStatus.setCachedVersion(appInfo.version.versionId, for: appInfo.id)
```
ts-leojoseph-r, line 177: These two lines should be moved before calling `isRequiredPermissionsAllowed()`, I guess this missed while doing conflict merge. Sorry about the confusion. 

The reason why we have moved them outside the block is to save the downloaded version details before checking `isRequiredPermissionsAllowed()`, so next time when create() is called, it will not download the same version again.

ts-leojoseph-r (commented):
Overall looks good. Waiting for the Mini App to test the full flow. Added few comments, please check.

Also, do you think we need to update the Changelog?

MiniApp/Classes/core/RealMiniApp.swift
Climbatize: That's good we have code reviews so you could catch it, thanks for spotting it ;)

Example/Controllers/Settings/AccessTokenViewController.swift
```diff
@@ -23,104 +20,43 @@ class AccessTokenViewController: UIViewController {
...
         self.accessTokenTextField.text = tokenInfo.tokenString
-        self.expiryDateTextField.text = formatDateToDisplay(date: tokenInfo.expiryDate)
+        self.datePicker.date = tokenInfo.expiryDate
     }
 
     func setDefaultTokenInfo() {
         self.accessTokenTextField.text = "ACCESS_TOKEN"
-        self.expiryDateTextField.text = formatDateToDisplay(date: Date())
-        _ = saveTokenInfo(accessToken: "ACCESS_TOKEN", expiryDate: Date())
+        self.datePicker.date = Date()
```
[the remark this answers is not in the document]
Climbatize: was my guess too. But maybe people will want to test expired tokens? @corycaywood your POV?

MiniApp/Classes/core/RealMiniApp.swift
Climbatize: we don't need anymore, I think this was my first implementation, and this did not appear during my refactoring

MiniApp/Classes/core/Protocols/MiniAppUserInfoDelegate.swift
```diff
@@ -31,18 +33,22 @@ public extension MiniAppUserInfoDelegate {
...
+    func getAccessToken(miniAppId: String, scopes: MASDKAccessTokenScopes?, completionHandler: @escaping (Result<MATokenInfo, MASDKCustomPermissionError>) -> Void) {
         completionHandler(.failure(.failedToConformToProtocol))
     }
 }
 
 public class MATokenInfo: Codable {
     let token: String
     let validUntil: Int
+    let audience: String?
+    let scopes: [String]?
```
corycaywood, line 45: See my above comment: I think these should be required properties.

Example/Controllers/Extensions/ViewController+MiniAppUserInfoProtocol.swift
```diff
@@ -23,11 +23,17 @@ extension ViewController {
...
+        var resultToken = "ACCESS_TOKEN"
+        var resultDate = Date()
+        let resultScopes = scopes
+
+        if let tokenInfo = getTokenInfo() {
+           resultToken = tokenInfo.tokenString
+           resultDate = tokenInfo.expiryDate
         }
-        completionHandler(.success(.init(accessToken: tokenInfo.tokenString, expirationDate: tokenInfo.expiryDate)))
+        completionHandler(.success(.init(accessToken: resultToken, expirationDate: resultDate, scopes: resultScopes)))
```
corycaywood, line 37: Looks like the JS Bridge code needs to be updated too. The Mini App currently isn't receiving the "scopes" because it's not parsed in the bridge code.

corycaywood (commented):
Looking good, just a few refinements I left below.

corycaywood (approved):
Looks good 👍 Just one issue with the public API.

MiniApp/Classes/core/Protocols/MiniAppUserInfoDelegate.swift
```diff
@@ -14,7 +14,9 @@ public protocol MiniAppUserInfoDelegate: class {
     func getContacts() -> [MAContact]?
 
     /// Interface that is used to retrieve the Token Info
-    func getAccessToken(miniAppId: String, completionHandler: @escaping (Result<MATokenInfo, MASDKCustomPermissionError>) -> Void)
```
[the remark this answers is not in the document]
Climbatize: oh, you're right

MiniApp/Classes/core/RealMiniApp.swift
ts-leojoseph-r: Looks like this change is missed @Climbatize

ts-leojoseph-r:
Can you update the ChangeLog as well?

ts-leojoseph-r (approved):
Looks good to me :+1:

PR #230, opened 2021-03-10
Compare previous and current manifest

ts-leojoseph-r:
# Description
* Compare `downloadedManifest` and the `latestManifest`. If any changes found in requiredPermissions, then display first-time screen to the user to agree the latest manifest.

## Links
[MINI-2509](https://jira.rakuten-it.com/jira/browse/MINI-2509)

# Checklist


MiniApp/Classes/core/Models/MiniAppManifest.swift
```diff
@@ -54,6 +54,10 @@ public struct MiniAppManifest: Codable {
         self.optionalPermissions = optionalPermissions
         self.customMetaData = customMetaData
     }
+
+    public static func == (lhs: MiniAppManifest, rhs: MiniAppManifest) -> Bool {
+        lhs.requiredPermissions == rhs.requiredPermissions
```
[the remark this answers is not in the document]
ts-leojoseph-r: Ok, Understood 👍 . Updating now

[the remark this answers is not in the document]
ts-leojoseph-r: Done

MiniApp/Classes/core/Models/MiniAppManifest.swift
```diff
@@ -54,6 +54,12 @@ public struct MiniAppManifest: Codable {
         self.optionalPermissions = optionalPermissions
         self.customMetaData = customMetaData
     }
+
+    public static func == (lhs: MiniAppManifest, rhs: MiniAppManifest) -> Bool {
+        lhs.requiredPermissions == rhs.requiredPermissions &&
+            lhs.optionalPermissions == rhs.optionalPermissions &&
+            lhs.customMetaData == rhs.customMetaData
```
Climbatize, line 61: you should compare `Set`s (`Set(lhs.requiredPermissions) == Set(rhs.requiredPermissions)`) or `.sorted()` your arrays to be sure 2 manifests with permissions ordered differently are equal

ts-leojoseph-r: Done

Example/Controllers/ViewController.swift
```diff
@@ -74,39 +74,99 @@ class ViewController: UIViewController {
...
+        let requiredPermissions = filterPermissions(permsArray: latestManifest.requiredPermissions ?? [],
+                                                    cachedPermissions: cachedAllowedPermissions)
+        let optionalPermissions = filterPermissions(permsArray: latestManifest.optionalPermissions ?? [],
+                                                    cachedPermissions: cachedPermissions)
+
+        if oldManifest == latestManifest && requiredPermissions.count == 0 {
+            self.displayMiniApp(miniAppInfo: miniAppInfo)
+        } else {
+            if requiredPermissions.count > 0 || optionalPermissions.count > 0 {
+                self.displayFirstTimeLaunchScreen(reqPermissions: requiredPermissions,
```
corycaywood, line 132: I think the first-time launch screen should display all of the permissions and meta-data again, not just the data that changed.

It might also be nice to label the new permissions with "New" or something, but that's something we can do later I think.

corycaywood (commented):
Looking good, but just one comment.

Also, could you add information about `getCurrentManifest` and how to compare manifests to the user guide?

Example/Controllers/ViewController.swift
ts-leojoseph-r: Do we really need to display all the permissions? Because the user has already responded to the permissions before right?
For eg., 1.0 version contains **Required** - Username :: **Optional** - Profile, Contact & Device. 

2.0 - **Required** - Username, Profile :: **Optional** - Contact & Device. 

Scenario 1: User agreed to all the permissions
There is a change in meta-data now, should we display all the permissions again?

Scenario 2: User didn't agree to **Profile** perm at 1.0 and launching 2.0 now
There is a change in meta-data now, should we display all the permissions again? or just the profile permission?

corycaywood: Yeah, I think we should show the same information on the launch screen in all of these cases. Basically show all information from the manifest: required permissions, optional permissions, and customMetaData.

This is because it could be confusing to the user that it's only showing the new permissions. It makes it seem like the Mini App only needs these new permissions and not the old permissions anymore.

The only difference is that maybe we can show a "New" label next to permissions that have been added (and an "Updated" label next to the `customMetaData` when it's changed)? But I think that's not necessary right now. We should keep it simple for now and can add an improvement for labels later.

corycaywood (commented):
Looks good! Just still need to update the user guide with information about `getCurrentManifest`. I think something similar to what's in the Android PR is okay: 
https://github.com/rakutentech/android-miniapp/pull/271/files#diff-e0502b4de6bd41029198132f0b09748d40b52a794167c81874134ff7a8021447R503

Example/Controllers/ViewController.swift
ts-leojoseph-r: > Looks good! Just still need to update the user guide with information about `getDownloadedManifest`. I think something similar to what's in the Android PR is okay:
> https://github.com/rakutentech/android-miniapp/pull/271/files#diff-e0502b4de6bd41029198132f0b09748d40b52a794167c81874134ff7a8021447R503

Done

corycaywood (approved):
Looks good! 👍

PR #231, opened 2021-03-11
Allow mini app to request only permissions that is defined in manifest

ts-leojoseph-r:
# Description
Ignore/filter to not request custom permissions if it is not mentioned in manifest

# Checklist

### DIFF
--- a/MiniApp/Classes/core/JavascriptBridge/MiniAppJavaScriptPermissionManager.swift
+++ b/MiniApp/Classes/core/JavascriptBridge/MiniAppJavaScriptPermissionManager.swift
@@ -1,6 +1,7 @@
 extension MiniAppScriptMessageHandler {
     func requestCustomPermissions(requestParam: RequestParameters?, callbackId: String) {
         cachedUnknownCustomPermissionRequest.removeAll()
+        permissionsNotAddedInManifest.removeAll()
         userAlreadyRespondedRequestList.removeAll()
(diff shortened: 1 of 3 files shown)


MiniApp/Classes/core/JavascriptBridge/MiniAppJavaScriptPermissionManager.swift
```diff
@@ -27,10 +27,15 @@ extension MiniAppScriptMessageHandler {
             miniAppPermissionRequestModelList.contains($0)
         }
 
-        let userNotRespondedRequestList = miniAppPermissionRequestModelList.filter {
+        var userNotRespondedRequestList = miniAppPermissionRequestModelList.filter {
             !allowedList.contains($0)
         }
 
+        /// To make sure that only permissions mentioned in the manifest is allowed at run-time
+        userNotRespondedRequestList = userNotRespondedRequestList.filter {
```
corycaywood, line 35: I think we need to return the same permission list that was passed by the user. Just if they pass a permission which isn't in the manifest, then it gets set to DENIED.

So I think we shouldn't remove any permissions, just set to DENIED if it's not in the manifest.

ts-leojoseph-r: Done

corycaywood (approved):
Looks good! 👍

Climbatize (approved):

PR #238, opened 2021-03-17
Refactor sample app and update Danger

ts-leojoseph-r:
# Description
* Updated Danger version
* Refactored `getConfig` -> `getCurrentConfig`

# Checklist


Climbatize (approved):
to me the dot (`.`) means `get` and `set`, so it should not be necessary to repeat it, e.g: `Config.getCurrent()` -> `Config.current()`. As for the change from `getCurrent()` to `getCurrentConfig()`, as we are in an object called Config, I feel it's redundant. But this is only my personal opinion, and this change only affects demo app so it does not bother me, so feel free to merge it :)

ts-leojoseph-r:
> to me the dot (`.`) means `get` and `set`, so it should not be necessary to repeat it, e.g: `Config.getCurrent()` -> `Config.current()`. As for the change from `getCurrent()` to `getCurrentConfig()`, as we are in an object called Config, I feel it's redundant. But this is only my personal opinion, and this change only affects demo app so it does not bother me, so feel free to merge it :)

I agree. For me, (`.`) means only `get` method, for `set` its better to add `set` prefix like apple does already. Update feedback changes

corycaywood (approved):
LGTM 👍

Climbatize (approved):
🙇‍♂️

PR #249, opened 2021-04-05
Mali 109 miniapp display UI

laubengaier:
# Description
MiniApp Display UI which allows simple usage `MiniApp.launch` to display a MiniApp.

## Links
https://jira.rakuten-it.com/jira/browse/MALI-109
https://confluence.rakuten-it.com/confluence/display/ONEAPP/Reusable+Mini+App+Screen+in+Link#ReusableMiniAppScreeninLink-HowdowemaketheMiniAppscreenre-usableinLink

# Checklist

First PR so feedback is very welcomed! 
For now It's not ultimately defined how it should look or work like so onError etc is not used yet.


MiniApp/Classes/core/RealMiniApp.swift
```diff
@@ -346,3 +346,19 @@ extension RealMiniApp: MiniAppMessageDelegate {
...
+// MARK: - UI
+extension RealMiniApp {
+    func launch(
+        base: UIViewController,
+        appId: String,
+        config: MiniAppSdkConfig? = nil,
+        messageInterface: MiniAppMessageDelegate? = nil,
+        queryParams: String? = nil
+    ) {
+        let miniAppVC = MiniAppViewController(appId: appId, config: config, messageDelegate: messageInterface ?? self, navDelegate: nil, queryParams: queryParams)
```
Climbatize, line 359: if you don't have the `MiniApp/UI` in your podfile this will not compile

MiniApp/Classes/ui/MiniAppViewController.swift
```diff
@@ -0,0 +1,241 @@
...
+        return view
+    }()
+
+    private lazy var forwardButton: UIBarButtonItem = {
+        let view = UIBarButtonItem(image: UIImage(named: "arrow_right")!, style: .plain, target: self, action: #selector(forwardPressed))
+        return view
+    }()
+
+    private lazy var closeButton: UIBarButtonItem = {
+        let view = UIBarButtonItem(title: "Close", style: .plain, target: self, action: #selector(closePressed))
```
Climbatize, line 33: Text should be localized

MiniApp/Classes/ui/MiniAppViewController.swift
```diff
@@ -0,0 +1,241 @@
...
+            ])
+        }
+
+        required init?(coder: NSCoder) { return nil }
+
+        @objc
+        func retryPressed() {
+            onRetry?()
+        }
+    }
```
Climbatize, line 240: maybe you could put this into a new file to have a smaller class

Climbatize (changes requested):
Seems good to me! Just some comments:

- Please execute a `pod deintegrate` before pushing ;)
- Also I see a big issue: you created a UI subspec, which includes a new controller. But you have references to this controller in the core subspec

MiniApp/Classes/core/RealMiniApp.swift
corycaywood: I have a couple of ideas about this:

- We could move the `MiniApp.launch` function into a new class within the `ui` folder like `MiniAppUi`.
- OR we could move the `MiniApp` and `RealMiniApp` extensions for `launch` into the `ui` folder (though I don't know whether this will actually work?)

What do you think?

laubengaier: For my part, I think both make sense. 
If I think of firebase for example they've split auth ui into a separate module so it's up to the user but I don't know if there's a case for MiniApp at the current state which benefits from the separation. But in any case separation is almost always cleaner so If we'll add more ui related stuff like eg. List display I'd vote for the first case MiniAppUI.
@Climbatize how about you?

MiniApp/Classes/core/MiniApp.swift
```diff
@@ -128,6 +128,25 @@ public class MiniApp: NSObject {
     }
 }
 
+// MARK: - UI
+public extension MiniApp {
+    func launch(
```
ts-leojoseph-r, line 133: Some documentation may be for the public API ?

MiniApp/Classes/core/RealMiniApp.swift
ts-leojoseph-r: As of now, we don't have too many UI dependencies from the Host app. And I see it's good to have a separate module just for the UI is a better option. I already see you had moved the files under the UI folder, But as @Climbatize said, we would have a compilation error if we don't have `MiniApp/UI` in podfile.

ts-leojoseph-r: May be moving extensions of launch into submodule would help 🤔

ts-leojoseph-r (commented):
👏 on your first PR in SDK.
Overall looks ok to me, except the submodule part which we are discussing already.
Also, it would be great to move most of the UI part of MiniAppViewController into `xib` or `storyboard`. Not now, maybe later.

MiniApp/Classes/core/RealMiniApp.swift
Climbatize: I think you can keep what you did by just moving your extensions to the `ui` folder. By the way having a MiniAppUI class does not bother me, both ways are viable.

MiniApp/Classes/core/MiniApp.swift
laubengaier: @ts-leojoseph-r thanks! sure I'll provide more documentation tomorrow. I've just pushed a commit with the changes related to moving the launch stuff to the ui component. 
For the UI part I'd suggest moving on from xibs or storyboards because imho coded ui is easier to read and much easier to maintain / compare etc #discuss :D. (@Climbatize what do you think)
_If SwiftUI will improve in the future I think also that's the way to go but not until apple fixes some stuff_

MiniApp/Classes/ui/MiniAppViewController.swift
```diff
@@ -1,7 +1,21 @@
 import Foundation
 import UIKit
 
-class MiniAppViewController: UIViewController {
+public protocol MiniAppUIDelegate: class {
+    func onError(error: Error)
+    func onSuccess()
+    func onForward()
+    func onBackward()
+    func onClose()
```
Climbatize, line 9: If I understood well, this delegate will be triggered by MiniAppViewController? If it's the case, when making a delegate I prefer using the following syntax: 
```swift
public protocol MiniAppUIDelegate: class {
    func miniApp(_ viewController: MiniAppViewController, didLaunchWith config: MiniAppSdkConfig? = nil, error: Error? = nil)
    func miniApp(_ viewController: MiniAppViewController, shouldExecute action: MiniAppNavigationAction)
}
```
Of course, you would have to add the `.close` action to `MiniAppNavigationAction`. But that's just my personal opinion. What do you think? @corycaywood, @ts-leojoseph-r

MiniApp/ja.lproj/Localizable.strings
```diff
@@ -2,6 +2,6 @@
 "Cancel_title" = "キャンセル";
 
 // MARK: - UI
-"ui_close_button_title" = "Close";
-"ui_loading_failed_error" = "An error has occured.\nPlease try again";
-"ui_retry_button_title" = "Retry";
+"ui_close_button_title" = "クローズ";
+"ui_loading_failed_error" = "情報を取得できませんでした。\n再度お試しください。";
```
Climbatize, line 6: we should make this text reviewed by another Japanese speaking person. The meaning is a bit different from the original message.

Climbatize (commented):
Nice improvement!. Some comments, nothing really serious ;)

MiniApp/ja.lproj/Localizable.strings
laubengaier: sure I've just copied it from Link except close 😄

MiniApp/Classes/ui/MiniAppViewController.swift
laubengaier: @Climbatize 
yeah + the delegate should be used when using the `launch` function. 
If the user wants to customize or take more stuff in their hands I've exposed a create function which allows to just override the vc and the functions without really need to implement the delegate.

Syntaxwise sure would allow easy access to the vc 👍

MiniApp/en.lproj/Localizable.strings
```diff
@@ -37,3 +37,8 @@
 "wait_message" = "Please wait...";
 "ok" = "OK";
 "cancel" = "Cancel";
+
+// MARK: - UI
+"ui_close_button_title" = "Close";
```
corycaywood, line 42: Are these the key names usable by the Host App to override the text values? Shouldn't they be prefixed with something specific to mini apps? i.e. `miniapp_ui_close_button_title`

MiniApp/Classes/ui/MiniAppUI.swift
```diff
@@ -0,0 +1,61 @@
...
+    public init(title: String = "", miniAppId: String, config: MiniAppSdkConfig? = nil, messageInterface: MiniAppMessageDelegate, queryParams: String? = nil) {
+        self.title = title
+        self.miniAppId = miniAppId
+        self.config = config
+        self.messageInterface = messageInterface
+        self.queryParams = queryParams
+    }
+}
+
+public class MiniAppUI {
```
corycaywood, line 20: Could you add documentation for this class and its public functions?

MiniApp/Classes/ui/MiniAppViewController.swift
```diff
@@ -0,0 +1,188 @@
...
+
+    let appId: String
+    var config: MiniAppSdkConfig?
+    var queryParams: String?
+    var state: ViewState = .loading {
+        didSet { update() }
+    }
+
+    weak var messageDelegate: MiniAppMessageDelegate?
+    weak var navDelegate: MiniAppNavigationDelegate?
```
corycaywood, line 28: There are some additional concerns about this we need to think about:

- We need to let the Host App provide their own `MiniAppNavigationDelegate` so that they can override certain links (i.e. deep links) or if there are certain links they want to handle differently.
- We need to also provide a default WebView for displaying links to external pages. The SDK already has a default WebView built-in, but it doesn't provide the full functionality because it needs access to the ViewController. Now, we have access to the ViewController here, so we can provide the full functionality (see the Example app code):
  - The Mini App has the capability to receive parameters back from the external page which was opened in a secondary WebView (and the external page has the capability to close the secondary WebView). This is done by handling a specific `miniapp://` scheme from the secondary WebView.

So basically I think we need to have two `MiniAppNavigationDelegate`: one which is provided by the Host App (and we check this one first to see if the Host App handles the link) and one which is a default in this ViewController.

corycaywood (commented):
It's looking really good. I left a couple comments below.

Also, could you update the Example App code to replace the current mini app ViewController with this one?

MiniApp/Classes/core/MiniApp.swift
laubengaier: @leorjoseph @Climbatize is it ok?

MiniApp/Classes/ui/MiniAppViewController.swift
```diff
@@ -0,0 +1,193 @@
...
+public protocol MiniAppUIDelegate: class {
+    func miniApp(_ viewController: MiniAppViewController, didLaunchWith config: MiniAppSdkConfig?)
+    func miniApp(_ viewController: MiniAppViewController, shouldExecute action: MiniAppNavigationAction)
+    func miniApp(_ viewController: MiniAppViewController, didLoadWith error: Error?)
+    func onClose()
+}
+public extension MiniAppUIDelegate {
+    func miniApp(_ viewController: MiniAppViewController, didLaunchWith config: MiniAppSdkConfig?) {}
+    func miniApp(_ viewController: MiniAppViewController, shouldExecute action: MiniAppNavigationAction) {}
+    func miniApp(_ viewController: MiniAppViewController, didLoadWith error: Error?) {}
```
ts-leojoseph-r, line 13: Default implementation for `onClose` missing or is that deliberate?

MiniApp/Classes/ui/MiniAppViewController.swift
```diff
@@ -0,0 +1,193 @@
...
+                self.activityIndicatorView.stopAnimating()
+                self.fallbackView.isHidden = true
+            }
+        }
+    }
+
+    @objc
+    public func backPressed() {
+        if delegate == nil {
+            delegate?.miniApp(self, shouldExecute: .back)
```
ts-leojoseph-r, line 169: 🤔 . Shouldn't this be !=

ts-leojoseph-r (commented):
Seems ok to me. Just few more comments to be addressed.
Please add documentation for all public interfaces/variables/protocols

MiniApp/Classes/ui/MiniAppViewController.swift
Climbatize: Good catch @ts-leojoseph-r might be right ^^. Maybe a `if let delegate = delegate` would remove the need to unwrap it :) I don't know what's the most elegant to be honest.
I'm wondering though, shouldn't we call both delegates if they are present?

laubengaier: ouh woops 😂

MiniApp/Classes/ui/MiniAppViewController.swift
laubengaier: @corycaywood which links exactly should I handle per default implementation?

MiniApp/ja.lproj/Localizable.strings
laubengaier: @Climbatize to whom can we reach out for the translations or should I just keep english for now?

corycaywood: I think we can keep just the english for now - as long as we make sure the strings can be localized by the Host App. But I think @Climbatize you will work on some changes to make strings in the SDK localizable?

Climbatize: yes that's true @corycaywood, I'll take in charge all these localization concerns

Climbatize (approved):
LGTM. Once you'll fix feedbacks from @corycaywood and @ts-leojoseph-r I think we have a good default implementation. Implementation of the navigation can be a little tricky so I hope we'll merge this PR soon and we will be able to test it for a long time and fix eventual wrong implementations while developing other features before the release. What do you think guys?

ts-leojoseph-r:
Looks ok to me. Just waiting for the CI to pass & approve. Looks like few test cases are failing.

MiniApp/ja.lproj/Localizable.strings
laubengaier: @Climbatize ok I'll comment out the japanese localization for now (using your localization imp later)

laubengaier:
@Climbatize @leorjoseph how are we dealing with code smells because it's eg showing errors with nested closures for the tests which I find kinda odd because it's working the same for other tests? how should I go on with that?

rleojoseph:
> @Climbatize @leorjoseph how are we dealing with code smells because it's eg showing errors with nested closures for the tests which I find kinda odd because it's working the same for other tests? how should I go on with that?

I just checked the code smell issues and they seem to be ok. Not a problem. As long as the Quality is passed from SonarCloud, we are good to go. We can however think of getting rid of them in the future.

laubengaier:
@Climbatize @leorjoseph is it ok to add this localization keys for the ui part?

MASDKLocale
```
case uiNavButtonClose                       = "miniapp.sdk.ios.ui.nav.button.close"
case uiFallbackTitle                        = "miniapp.sdk.ios.ui.fallback.title"
case uiFallbackButtonRetry                  = "miniapp.sdk.ios.ui.fallback.button.retry"
```

Localizable.strings
```
/* MASDKUI */
"miniapp.sdk.ios.ui.nav.button.close"                           = "Close"
"miniapp.sdk.ios.ui.fallback.title"                             = "An error has occured.\nPlease try again"
"miniapp.sdk.ios.ui.fallback.button.retry"                      = "Retry"
```

corycaywood (approved):
Looks good 👍  

But QA won't be able to test this currently - I think we should update the Example App to replace the current UI with this built-in UI. But maybe you can do that in a separate PR?

laubengaier:
> Looks good 👍
> 
> But QA won't be able to test this currently - I think we should update the Example App to replace the current UI with this built-in UI. But maybe you can do that in a separate PR?

@corycaywood thanks, sure I can do that. do you know if the localization thing is appropriate to do so?

rleojoseph:
> @Climbatize @rleojoseph is it ok to add this localization keys for the ui part?

I guess it's fine. @Climbatize already added one in the SDK and the below localizable looks ok to be added.

> and is it possible to somehow load an example miniapp for the unit tests? _It's probably failing now because the config is empty with MiniAppSdkConfig() not loading any example project._

That's right, the config would be nil. We don't have any testcases written so far to launch a mini-app, however we added UT's for checking whether mini-app view is returned properly using `getMiniAppView()`

PR #253, opened 2021-04-07
Caching MiniAppManifest response

ts-leojoseph-r:
# Description
* Cache Mini App manifest response to avoid multiple API request for the same ID+Version 

## Links
MINI-2944

# Checklist


MiniApp/Classes/core/MiniAppInfoFetcher.swift
```diff
@@ -44,21 +50,31 @@ internal class MiniAppInfoFetcher {
 
     func getMiniAppMetaInfo(miniAppId: String, miniAppVersion: String, apiClient: MiniAppClient, completionHandler: @escaping (Result<MiniAppManifest, MASDKError>) -> Void) {
 
-        apiClient.getMiniAppMetaData(appId: miniAppId, versionId: miniAppVersion) { (result) in
-            switch result {
-            case .success(let responseData):
-                guard let decodeResponse = ResponseDecoder.decode(decodeType: MetaDataResponse.self,
-                    data: responseData.data) else {
-                    return completionHandler(.failure(.invalidResponseData))
+        guard let manifest = getCachedManifest(miniAppId: miniAppId, miniAppVersion: miniAppVersion) else {
```
corycaywood, line 53: I don't think this is the right place handle the caching. This caching is being added because we want to avoid making multiple API calls for the same thing twice in row, so I think the caching should be added closer to the API functionality - i.e. in [`ManifestDownloader`](https://github.com/rakutentech/ios-miniapp/blob/master/MiniApp/Classes/core/ManifestDownloader.swift).

So basically `ManifestDownloader` will cache just a single manifest (the last one which was fetched) and also which mini app id/version it is for. So the next time that a request is made to `ManifestDownloader`, it will check whether it's the same Manifest which was just fetched, and, if so, then immediately return the cached manifest.

MiniApp/Classes/core/Storage/MAManifestCache.swift
```diff
@@ -0,0 +1,59 @@
...
+        }
+        guard let allKeys = retrieveAllManifestInfo(), let manifestInfo = allKeys[appId + "/" + versionId] as CachedMetaData? else {
+            return nil
+        }
+        return manifestInfo
+    }
+
+    /// Remove Key from the KeyChain
+    /// - Parameter keyId: Mini app ID
+    internal func removeKey(forMiniApp appId: String, versionId: String) {
```
corycaywood, line 37: Currently no keys are ever being removed, right? So we could end up with many cached manifests that we don't need.

As mentioned in the comment above, I think we only need to cache a single manifest (the latest fetched), so probably this cache class can be simplified.

corycaywood (commented):
The basic idea is good, but I think it needs to be refactored into the `ManifestDownloader`. See the Android implementation:
- https://github.com/rakutentech/android-miniapp/blob/master/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt#L134
- https://github.com/rakutentech/android-miniapp/blob/master/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ManifestApiCache.kt

corycaywood:
Alternatively, is there any way to enable caching at the HTTP layer? Though I'm not sure if this will work unless the API response has the correct headers.

MiniApp/Classes/core/MiniAppInfoFetcher.swift
Climbatize: > we want to avoid making multiple API calls for the same thing twice in row

does this mean we would only need to store it in a variable of the MiniApp singleton?

ts-leojoseph-r: @corycaywood [ManifestDownloader](https://github.com/rakutentech/ios-miniapp/blob/master/MiniApp/Classes/core/ManifestDownloader.swift) is meant for list of manifest files to download for the mini app. But this ticket is meant to cache the meta-data right ?

ts-leojoseph-r: @corycaywood [ManifestDownloader](https://github.com/rakutentech/ios-miniapp/blob/master/MiniApp/Classes/core/ManifestDownloader.swift) is meant for list of manifest files to be downloaded for MiniApp. But this ticket is for meta-data caching right ?

corycaywood: @leorjoseph Oh sorry, yeah that's right. Then maybe we need to create a `MetaDataDownloader` class to encapsulate this functionality?

@Climbatize Yeah, I guess we could just cache the response in memory. But the manifest.json could be very long for some mini apps, so I'm not sure whether we should do that. What do you think?

ts-leojoseph-r: Done

ts-leojoseph-r:
> Alternatively, is there any way to enable caching at the HTTP layer? Though I'm not sure if this will work unless the API response has the correct headers.

The initial plan which I had in my mind is use the [HTTP layer caching](https://developer.apple.com/documentation/foundation/nsurlrequest/cachepolicy/useprotocolcachepolicy). But like you said, our API isn't having the right headers now.

corycaywood (approved):
Looks good! 👍

Climbatize (approved):
Have my thumb up

PR #270, opened 2021-05-13
Fix MINI-3237

rleojoseph:
# Description
* Sometimes host app fails to present a new controller immediately after the user responds to the Custom Permission screen. Like in MINI-3237. This forces the host app to display a new controller using DispatchQueue/Delay.

So we are sending back the Custom permission response after the view is dismissed. There is no noticable delay

## Links
MINI-3237

# Checklist


Climbatize (approved):
I missed this on the review, better this way ;)

corycaywood (approved):
👍

PR #271, opened 2021-05-18
Update getAccessToken to support new error types

rleojoseph:
# Description
Updated `getAccessToken` method to return new error type `MASDKAccessTokenError`

`MASDKAccessTokenError` will support following errors,

* FailedToConformToProtocol
* AudienceNotSupportedError
* ScopesNotSupportedError
* AuthorizationFailureError
* Error

## Links
MINI-3132

# Checklist


MiniApp/Classes/core/Protocols/MiniAppMessageDelegate.swift
```diff
@@ -150,3 +150,54 @@ public enum MASDKCustomPermissionError: String, MiniAppErrorProtocol {
...
+        switch self {
+        case .failedToConformToProtocol:
+            return MASDKLocale.localize(.failedToConformToProtocol)
+        case .audienceNotSupportedError:
+            return "The value passed for 'audience' is not supported."
+        case .scopesNotSupportedError:
+            return "The value passed for 'scopes' is not supported."
+        case .authorizationFailureError(let description):
+            return description
+        case .error(let description):
```
Climbatize, line 183: in JS SDK we called this kind of errors "Other", do you think we can bring this to our own SDK?

MiniApp/Classes/core/Protocols/MiniAppMessageDelegate.swift
```diff
@@ -150,3 +150,54 @@ public enum MASDKCustomPermissionError: String, MiniAppErrorProtocol {
...
+            return description
+        }
+    }
+
+    /// Title of the error
+    public var name: String {
```
Climbatize, line 189: maybe we could rename this one `rawValue` to mimic a string enum behavior?

Climbatize (approved):
LGTM, just some trivial comments

MiniApp/Classes/core/Protocols/MiniAppMessageDelegate.swift
rleojoseph: Yeah, not a problem. But I sort of followed the confluence documentation. Should we update the documentation as well ?

MiniApp/Classes/core/Protocols/MiniAppMessageDelegate.swift
rleojoseph: No, these error is conformed to `MiniAppErrorProtocol` which has `name` & `description`. We can update them to `rawValue` but that involves in updating the `MiniAppErrorProtocol` itself. Is that okay ?

laubengaier (approved):
LGTM! 👍

MiniApp/Classes/core/Protocols/MiniAppMessageDelegate.swift
Climbatize: no that's fine, thanks for the heads up ;)

MiniApp/Classes/core/Protocols/MiniAppMessageDelegate.swift
Climbatize: I think that's not that important, as what I did on JS side is that if I receive an unknown error I set is as `Other` type. We can go with this implementation.

PR #273, opened 2021-05-25
localizable crash fix

Climbatize:
# Description
variadic methods to manage string formats are working well, but unfortunately if the localizable string format is not well formatted the resulting crash is difficult to investigate. So I suggest to deprecate them and let the developer use the String(format:_:) method himself, as you can see in this pull request

# Checklist


rleojoseph (approved):
Looks good to me :+1:

PR #280, opened 2021-06-09
added QA settings screen

Climbatize:
# Description
This PR adds a QA settings screen for particular needs during test

https://user-images.githubusercontent.com/1507391/121446654-f5402600-c9ce-11eb-8f7c-ef150f964432.mp4

## Links
-
# Checklist


rleojoseph (approved):
Good job providing a quick UI for this approach 👏🏻 . Also, I feel the same video should be shared with the QA, it is self-descriptive.

One suggestion: Is it possible to move the Access Token error page one more level down? Like Settings-> QA-> Access Token-> Access Token error items. I believe more things will be added in the QA section, that's why.
 
Looks good to me :+1:

Climbatize:
> Good job providing a quick UI for this approach 👏🏻 . Also, I feel the same video should be shared with the QA, it is self-descriptive.

I agree we should share the video, thanks for the positive review 👍 

> One suggestion: Is it possible to move the Access Token error page one more level down? Like Settings-> QA-> Access Token-> Access Token error items. I believe more things will be added in the QA section, that's why.
> 
I had that in mind when I created the screen, that's why for now the QA options are regrouped by grouped sections in the static tableView. So in order to avoid more manipulation I suggest we keep this as simple as it is until it becomes a real mess, what do you think @rleojoseph?

rleojoseph:
> > One suggestion: Is it possible to move the Access Token error page one more level down? Like Settings-> QA-> Access Token-> Access Token error items. I believe more things will be added in the QA section, that's why.
> 
> I had that in mind when I created the screen, that's why for now the QA options are regrouped by grouped sections in the static tableView. So in order to avoid more manipulation I suggest we keep this as simple as it is until it becomes a real mess, what do you think @rleojoseph?

That's fine for now

khairul-alam-licon (approved):
👍👍

PR #286, opened 2021-06-15
Updated Additional RAT account in Demo app.

rleojoseph:
# Description
RAT is not working due to the default value set in the PLIST.
Now it is updated, also added another RAT account to send analytics.

## Links
Add links to github/jira issues, design documents and other relevant resources (e.g. previous discussions, platform/tool documentation etc.)

# Checklist


Climbatize (approved):
LGTM

khairul-alam-licon (approved):
👍👍

Example/Utils/Config.swift
```diff
@@ -13,12 +13,15 @@ class Config: NSObject {
...
+    // swiftlint:disable:next todo
+    // TODO: Make it as CI Configurable
     class func current() -> MiniAppSdkConfig {
         MiniAppSdkConfig(baseUrl: Config.userDefaults?.string(forKey: Config.Key.endpoint.rawValue),
             rasProjectId: Config.userDefaults?.string(forKey: Config.Key.projectId.rawValue),
             subscriptionKey: Config.userDefaults?.string(forKey: Config.Key.subscriptionKey.rawValue),
             hostAppVersion: Config.userDefaults?.string(forKey: Config.Key.version.rawValue),
-            isPreviewMode: Config.userDefaults?.bool(forKey: Config.Key.isPreviewMode.rawValue))
+            isPreviewMode: Config.userDefaults?.bool(forKey: Config.Key.isPreviewMode.rawValue),
+            analyticsConfigList: [MAAnalyticsConfig(acc: "477", aid: "998")])
```
munir-rakuten, line 24: `acc` `aid` values can be stored in a global variable or env file, just a suggestion 🙂

munir-rakuten (approved):
Looks good to me  👍

Example/Utils/Config.swift
rleojoseph: @munir-rakuten , this is a quick fix. Also, we are planning to make changes to this by next week (from Settings). So, I felt it is ok for now

corycaywood (approved):

Example/Utils/Config.swift
donnie-jp: acc and aid values are internal values that we shouldn't be exposing in source code. That was why I made them configurable from the plist

rleojoseph: @donnie-jp . We have the default details hardcoded already in the SDK's code already https://github.com/rakutentech/ios-miniapp/blob/master/MiniApp/Classes/core/MiniAppAnalytics.swift#L52. Thought this was ok with the additional account, do you want me to remove it immediately?

donnie-jp: I guess it's up to your team / @corycaywood. Just flagging the issue cos I happened to see the PR title on bitrise.

corycaywood: @donnie-jp We will later be replacing these values with a setting in the Demo App UI.

As for the `1553` setting used by the SDK: this value needs to always be used by the SDK and we don't want the Host App to be able to change it. So as far as I understand, there isn't a way to set it from an environment variable.

PR #305, opened 2021-07-19
release candidate script

Climbatize:
# Description
I made this script to semi-automate Release Candidate branch creation

### DIFF
--- a/scripts/release-candidate.sh
+++ b/scripts/release-candidate.sh
@@ -10,7 +10,7 @@ echo "
         -b Branch       Branch to release. By default 'candidate'
         -d              displays useful data to debug this script
         -a              automatic mode. Requires -v parameter to be 100% without prompt
-        -s              silent mode
+        -s              same as -a but in silent mode
 
         For Example: ./release-candidate.sh -v 3.5.0 -b candidate -a
 
@@ -18,7 +18,6 @@ echo "
 "
 }
 NO_PROMPT=0
-SILENT=0
 
 while getopts ":v:b:dhas" opt; do
   case $opt in
@@ -32,7 +31,7 @@ while getopts ":v:b:dhas" opt; do
     ;;
     a) NO_PROMPT=1
     ;;
-    s) SILENT=1; echo off
+    s) NO_PROMPT=1; exec &>/dev/null
     ;;
     \?) echo "Invalid option -$OPTARG" >&2; exit 1
     ;;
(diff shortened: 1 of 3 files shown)


scripts/release-candidate.sh
```diff
@@ -0,0 +1,58 @@
...
+    echo "Can't continue execution. Proceed manually"
+    exit 1
+fi
+
+echo "Committing candidate version number changes"
+git commit -a -m "v$VERSION candidate"
+git push upstream_release "$CANDIDATE_BRANCH":candidate --force
+git checkout master
+git branch -D "$CANDIDATE_BRANCH"
+git remote remove upstream_release
```
corycaywood, line 56: Instead of adding the remote and then removing it, you could instead create a variable for the remote URL and just reference that directly. For example:

```
UPSTREAM_RELEASE=git@github.com:rakutentech/ios-miniapp.git
// ...
git fetch $UPSTREAM_RELEASE master:"$CANDIDATE_BRANCH"
git checkout "$CANDIDATE_BRANCH"
```

corycaywood (commented):
Nice idea!

Could you also add instructions about this in the dev readme?

scripts/release-candidate.sh
Climbatize: that's a great idea!

rleojoseph (approved):
Great Job 👏🏻 . It's gonna save us some time.
Looks good to me :+1:

scripts/release-candidate.sh
```diff
@@ -0,0 +1,115 @@
...
+  read -r -p "Is CHANGELOG.md up to date? Input Y after you think it is, or N to abort release[Y/n]: " answer
+
+  if [ "$answer" != "" ] && [ "$answer" == "${answer#[Nn]}" ] ;then
+      echo "Can't continue execution. Proceed manually"
+      exit 1
+  fi
+fi
+
+echo "Committing candidate version number changes"
+git commit -a -m "v$VERSION candidate"
```
corycaywood, line 109: This is actually updating the version for the release, not just the candidate. So maybe the commit message should be something like `build: Update podspec to v$VERSION`?

Also, it seems a little dangerous to use `-a`. Could you instead just use `git add MiniApp.podspec`?

Climbatize: absolutely

khairul-alam-licon (approved):
👍👍

PR #320, opened 2021-08-10
Updated code to support Enable/Disable Nav buttons  in MiniAppUI

rleojoseph:
# Description
* Initial support for this behavior is added in this [Pull Request](https://github.com/rakutentech/ios-miniapp/pull/309), but looks like there is a recent change on how we launch the Mini App from the sample app.
* Code is updated with `MiniAppUI.shared().launch()` so we have to add support for Enabling/Disabling buttons for this one as well, hence the PR.

## Links
[MINI-3351](https://jira.rakuten-it.com/jira/browse/MINI-3351)

# Checklist

### DIFF
--- a/MiniApp/Classes/ui/MiniAppViewController.swift
+++ b/MiniApp/Classes/ui/MiniAppViewController.swift
@@ -36,7 +36,7 @@ public class MiniAppViewController: UIViewController {
     weak var navDelegate: MiniAppNavigationDelegate?
     weak var navBarDelegate: MiniAppNavigationBarDelegate?
 
-    weak var delegate: MiniAppUIDelegate?
+    weak var miniAppUiDelegate: MiniAppUIDelegate?
 
     // MARK: UI - Navigation
     private lazy var backButton: UIBarButtonItem = {
@@ -154,10 +154,10 @@ public class MiniAppViewController: UIViewController {
                         view.frame = self.view.bounds
                         self.view.addSubview(view)
                         self.navBarDelegate = miniAppDisplay as? MiniAppNavigationBarDelegate
-                        self.delegate?.miniApp(self, didLoadWith: nil)
+                        self.miniAppUiDelegate?.miniApp(self, didLoadWith: nil)
                         self.state = .success
                     case .failure(let error):
-                        self.delegate?.miniApp(self, didLoadWith: error)
+                        self.miniAppUiDelegate?.miniApp(self, didLoadWith: error)
                         self.state = .error
                     }
                 },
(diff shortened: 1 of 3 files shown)


MiniApp/Classes/ui/MiniAppViewController.swift
```diff
@@ -191,31 +191,27 @@ public class MiniAppViewController: UIViewController {
 
     @objc
     public func backPressed() {
-        if delegate == nil {
-            navBarDelegate?.miniAppNavigationBar(didTriggerAction: .back)
-        } else {
-            delegate?.miniApp(self, shouldExecute: .back)
-        }
```
rleojoseph, line 198: @laubengaier I removed these lines because, I don't see any reason why we call `delegate?.miniApp(self, shouldExecute: .back)` here. Because the host app neither returns anything (Bool value to execute or not) nor host app doesn't have instance of `navigationBarDelegate` to perform the trigger to go `.back` & `.forward`.

Checking with you here so that removing this won't impact anything on the MALI side.

Climbatize: @rleojoseph if I remember well, when I created the navigation delegates for the first time I gave the ability to host app developers to have their own navigation logic, so the possibility to have no navigation bar. But as @laubengaier  made improvements to it maybe now we don't need this anymore

Climbatize (approved):
As long as @laubengaier is ok with it, LGTM 👍

MiniApp/Classes/ui/MiniAppViewController.swift
rleojoseph: @Climbatize , we still can use your delegates as intended. But these delegates are created/meant for a different reason I guess. Lets hear from @laubengaier

corycaywood (approved):
👍

MiniApp/Classes/ui/MiniAppViewController.swift
laubengaier: @leorjoseph yeah as far as I remember the purpose was to allow to override the back and forward actions but if the state is correctly passed and can be accessed via a delegate I see no issue. 
Just a question it's not necessary to provide a (navigation) view for the navigationDelegate to work right?

laubengaier (approved):
lgtm just a comment and a question to the discussion

rleojoseph:
> lgtm just a comment and a question to the discussion

@laubengaier Thanks. I don't see the question, what is it ?

MiniApp/Classes/ui/MiniAppViewController.swift
laubengaier: > > lgtm just a comment and a question to the discussion
> 
> @laubengaier Thanks. I don't see the question, what is it ?

damn, must have been accidentally deleted.
If I remember correctly this behavior was indeed as Pierre described to let the host app have their own navigation logic or react to forward / backward changes. If the state is correctly passed through the navigation delegate I think it's fine.
The question was if the navigation delegate is working without the (navigation) view? I remember briefly that something was not working as expected.
_But anyways I have implemented a workaround for the issue we had so shouldn't be critical_

rleojoseph: @laubengaier 
We still support that I believe. Host app can still receive them and can control whether to go back or forward. 
And like you said, yeah there is a problem with `MiniAppNavigationDelegate` here. 
When you create & launch the MiniApp using `MiniAppUI.shared().launch()` you still have the `MiniAppNavigationDelegate` referred/infered to Host app ViewController, and it works like below,

observeWebView()  	- SDK
	|
	|
miniAppNavigationCanGo(back: Bool, forward: Bool) {}.	- Hostapp
	|
	|
 refreshNavigationBarButtons(backButtonEnabled: Bool, forwardButtonEnabled: Bool) {} - SDK UI

PR #330, opened 2021-09-08
[MINI-3419] Signature verification

Climbatize:
# Description
This PR implements files signature check in Mini App SDK. It is partly based on @imaginaris work in https://github.com/rakutentech/ios-signatureverifier, the SDK code has been migrated and softly refactored to better suit our needs.
Implementing signature check Pod automatically enables it in MiniApp SDK. But by default a signature check failure does not prevent the Mini App to be displayed, it just triggers an analytics event. An optional parameter has been added in the case we would want to prevent it.
 
A setting screen has been added to help testing signature check failure parameter.

Screen Shot 2021-09-08 at 9 34 36 
Screen Shot 2021-09-08 at 9 34 47

## Links
[MINI-3419](https://jira.rakuten-it.com/jira/browse/MINI-3419)

# Checklist


laubengaier (approved):
nice work, left a little regex improvement 😄

MiniApp/Classes/signature/Security/SignatureKeyStore.swift
```diff
@@ -0,0 +1,88 @@
...
+                kSecAttrAccount as String: account,
+                kSecValueData as String: data,
+                kSecAttrAccessible as String: kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly
+            ]
+            status = SecItemAdd(queryAdd as CFDictionary, nil)
+        }
+
+        if status != errSecSuccess {
+            let error: String?
+            if #available(iOS 11.3, *) {
```
imaginaris, line 57: I think this #if can be removed because your target version is iOS 13.0, right?

Climbatize: Of course! Thanks for the heads up

MiniApp/Classes/core/Downloader/MiniAppClient.swift
```diff
@@ -145,7 +161,22 @@ internal class MiniAppClient: NSObject, URLSessionDownloadDelegate {
...
+                    self?.delegate?.fileDownloaded(at: location, downloadedURL: destinationURL, signatureChecked: isVerified || !requireMiniAppSignatureVerification)
+                }
+            } else {
+                MiniAppAnalytics.sendAnalytics(event: .signatureFailure, miniAppId: ids.0, miniAppVersion: ids.1)
+                delegate?.fileDownloaded(at: location, downloadedURL: destinationURL, signatureChecked: !requireMiniAppSignatureVerification)
+            }
+        #else
+            MiniAppAnalytics.sendAnalytics(event: .signatureFailure, miniAppId: ids.0, miniAppVersion: ids.1)
+            delegate?.fileDownloaded(at: location, downloadedURL: destinationURL)
+        #endif
```
rleojoseph, line 179: Do you think we can movubg this to a separate method would be good idea?

MiniApp/Classes/core/MiniApp.swift
```diff
@@ -1,7 +1,7 @@
 /// Mini App Public API methods
 public class MiniApp: NSObject {
     private static let shared = MiniApp()
-    private let realMiniApp = RealMiniApp()
+    let realMiniApp = RealMiniApp()
```
rleojoseph, line 4: Any reason? I see the build getting through even with private.

rleojoseph (approved):
Overall, Looks good to me :+1:

Just a few comments.

MiniApp/Classes/core/MiniAppAnalytics.swift
```diff
@@ -10,13 +10,16 @@ internal enum MiniAppRATEvent: String, CaseIterable {
...
 
     func name() -> String {
-        return "mini_app_\(rawValue)"
+        "mini_app_\(rawValue)"
     }
 
     func eType() -> MiniAppRATEventType {
         switch self {
+        case .signatureFailure:
+            return .custom
```
corycaywood, line 22: I don't think this is correct - we can only use pre-defined values for the `etype`. There aren't really any that make sense for this case, but maybe `async` is the closest? And then we would need to set another parameter to distinguish that this is the signature failure event. How about this?

```
etype = async
actype = signature_failure
```

Example/Controllers/Extensions/ViewController+MiniApp.swift
```diff
@@ -112,6 +112,10 @@ extension ViewController: MiniAppNavigationDelegate {
         switch error {
         case .metaDataFailure:
             metaDataFailure()
+        case .invalidSignature:
+            self.displayAlert(title: MASDKLocale.localize("miniapp.sdk.ios.error.title"), message: MASDKLocale.localize(.signatureFailed), dismissController: true) { _ in
```
corycaywood, line 116: Does App Center let you manually send a crash report/log? Maybe we could do that here so we have that extra information from the Demo App?

corycaywood (approved):
Nice work! 👍  Just see my comment about analytics.

MiniApp/Classes/core/MiniAppAnalytics.swift
corycaywood: I just noticed that the following is what @khairul-alam-licon  used on Android, so probably this is okay:
```
etype = click
actype = mini_app_signature_validation_fail
```

MiniApp/Classes/core/Downloader/MiniAppClient.swift
Climbatize: yeah, makes things cleaner

MiniApp/Classes/core/MiniApp.swift
Climbatize: it was for some tests I made and did not need in the end, I'll fix this

Example/Controllers/Extensions/ViewController+MiniApp.swift
Climbatize: Unfortunately App Center does not have this feature, we could have a dedicated slack channel for this kind of things

PR #342, opened 2021-10-19
[MINI-1976] MiniApp custom events

Climbatize:
# Description

- Implemented custom event mechanism (for now: pause, resume and webview closed)
- Duplicates JS logs into Swift console for a better debugging (DEBUG mode)

pause/resume events partly rely on notification Center due to the variety of different reasons that lead a MiniApp to appear/disappear.

## Links
[MINI-1976](https://jira.rakuten-it.com/jira/browse/MINI-1976)

# Checklist

### DIFF
--- a/MiniApp/Classes/core/Assets/console.js
+++ b/MiniApp/Classes/core/Assets/console.js
@@ -1,0 +1,22 @@
+function log(emoji, type, args) {
+  window.webkit.messageHandlers.logging.postMessage(
+    `${emoji} console.${type}: ${Object.values(args)
+      .map(v => typeof(v) === "undefined" ? "undefined" : typeof(v) === "object" ? JSON.stringify(v) : v.toString())
+      .map(v => v.substring(0, 3000)) // Limit msg to 3000 chars
+      .join(", ")}`
+  )
+}
+
+let originalLog = console.log
+let originalWarn = console.warn
+let originalError = console.error
+let originalDebug = console.debug
+
+console.log = function() { log("📗", "log", arguments); originalLog.apply(null, arguments) }
+console.warn = function() { log("📙", "warning", arguments); originalWarn.apply(null, arguments) }
+console.error = function() { log("📕", "error", arguments); originalError.apply(null, arguments) }
+console.debug = function() { log("📘", "debug", arguments); originalDebug.apply(null, arguments) }
+
+window.addEventListener("error", function(e) {
+   log("💥", "Uncaught", [`${e.message} at ${e.filename}:${e.lineno}:${e.colno}`])
+})
--- a/MiniApp/Classes/core/Extensions/WKUserContentController+MiniApp.swift
+++ b/MiniApp/Classes/core/Extensions/WKUserContentController+MiniApp.swift
@@ -10,25 +10,37 @@ extension WKUserContentController {
     ///   - miniAppId: Mini App id String value
     ///   - miniAppTitle: Mini App title provided to Javascript handler
     func addMiniAppScriptMessageHandler(delegate: MiniAppCallbackDelegate, hostAppMessageDelegate: MiniAppMessageDelegate, adsDisplayer: MiniAppAdDisplayer?, miniAppId: String, miniAppTitle: String) {
-        add(MiniAppScriptMessageHandler(
-                delegate: delegate,
-                hostAppMessageDelegate: hostAppMessageDelegate,
-                adsDisplayer: adsDisplayer,
-                miniAppId: miniAppId,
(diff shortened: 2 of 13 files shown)


MiniApp/Classes/core/Assets/console.js
```diff
@@ -0,0 +1,22 @@
...
+
+let originalLog = console.log
+let originalWarn = console.warn
+let originalError = console.error
+let originalDebug = console.debug
+
+console.log = function() { log("📗", "log", arguments); originalLog.apply(null, arguments) }
+console.warn = function() { log("📙", "warning", arguments); originalWarn.apply(null, arguments) }
+console.error = function() { log("📕", "error", arguments); originalError.apply(null, arguments) }
+console.debug = function() { log("📘", "debug", arguments); originalDebug.apply(null, arguments) }
```
corycaywood, line 18: I think this should be moved to the `js-miniapp-bridge` code so that it can be covered by unit tests. The `ios-bridge.ts` file is only included for iOS, so you can add it there without affecting Android:
https://github.com/rakutentech/js-miniapp/blob/master/js-miniapp-bridge/src/ios/ios-bridge.ts

MiniApp/Classes/core/Assets/console.js
```diff
@@ -0,0 +1,22 @@
...
+
+window.addEventListener("error", function(e) {
+   log("💥", "Uncaught", [`${e.message} at ${e.filename}:${e.lineno}:${e.colno}`])
+})
```
corycaywood, line 22: I think this listener should be `miniapperror` - `error` is too generic so this could conflict with other listeners in used by the mini app.

Also, just to make sure - this will only be available when the Host App is built as debug, right? The end developers of Mini Apps won't see these logs?

MiniApp/Classes/core/Utilities/Constants.swift
```diff
@@ -1,9 +1,13 @@
-struct Constants {
+enum Constants {
     static let miniAppSchemePrefix = "mscheme."
     static let rootFileName = "index.html"
-    static let javascriptInterfaceName = "MiniAppiOS"
-    static let javascriptSuccessCallback = "MiniAppBridge.execSuccessCallback"
-    static let javascriptErrorCallback = "MiniAppBridge.execErrorCallback"
+    enum JavaScript {
+        static let interfaceName = "MiniAppiOS"
+        static let logHandler = "logging"
```
corycaywood, line 6: Should this name be more specific to avoid conflicts? i.e. `MiniAppLogging`?

MiniApp/Classes/core/Assets/console.js
Climbatize: yep @corycaywood , this script will only be included when the pod (thus the host app) is compiled in debug mode

MiniApp/Classes/ui/MiniAppViewController.swift
```diff
@@ -32,6 +32,16 @@ public class MiniAppViewController: UIViewController {
...
     }
 
+    public override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        NotificationCenter.default.sendCustomEvent(MiniAppEvent.Event(type: .resume, comment: "MiniApp view did appear"))
+    }
+
+    public override func viewWillDisappear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
+        NotificationCenter.default.sendCustomEvent(MiniAppEvent.Event(type: .resume, comment: "MiniApp view will disappear"))
```
rleojoseph, line 42: This is supposed to a `.pause` event right ??

MiniApp/Classes/core/Extensions/WKUserContentController+MiniApp.swift
```diff
@@ -10,25 +10,37 @@ extension WKUserContentController {
...
+                    miniAppId: miniAppId,
+                    miniAppTitle: miniAppTitle
+            ), name: name)
+        }
     }
 
     /// Method to remove the custom WKScriptMessageHandler. Removing the message handler will help to avoid memory leaks
     func removeMessageHandler() {
-        removeScriptMessageHandler(forName: Constants.javascriptInterfaceName)
+        removeScriptMessageHandler(forName: Constants.JavaScript.interfaceName)
```
rleojoseph, line 27: You need to remove script handler for `.logHandler` as well

MiniApp/Classes/ui/MiniAppViewController.swift
Climbatize: nice catch

PR #356, opened 2021-11-22
[MINI-4369] Switch Staging & Production Environment

laubengaier:
# Description
Implementing Switch for Staging and Production Environment

https://user-images.githubusercontent.com/17570451/142817400-501e41e5-eae9-403d-84be-45f75e138ca5.mp4

## Links
https://jira.rakuten-it.com/jira/browse/MINI-4360

# Checklist


corycaywood (approved):
Looks good 👍

Example/Utils/Config.swift
```diff
@@ -8,30 +8,46 @@ class Config: NSObject {
...
              sslKeyHash                          = "RMASSLKeyHash"
     }
 
-    static let userDefaults = UserDefaults(suiteName: "com.rakuten.tech.mobile.miniapp.MiniAppDemo.settings")
+    private enum InternalKey: String {
+        case stagingActive = "MiniApp.Settings.StagingEnvironmentActive"
+    }
+
+    static var isProd: Bool {
+        !UserDefaults.standard.bool(forKey: InternalKey.stagingActive.rawValue)
```
Climbatize, line 24: This might crash at first launch

```suggestion
        if let prod = UserDefaults.standard.bool(forKey: InternalKey.stagingActive.rawValue) {
            return prod
        } else {
            return false
        }
```

Climbatize (changes requested):
LGTM except for the suggestion I provided. Also, before merging this one warn the team so we integrate the secrets in our environment ;)

Example/Utils/Config.swift
laubengaier: @Climbatize 
Can you explain more in detail why it should crash? 
UserDefaults returns false for a non existing key so it behaves as expected using prod first. I've tested couple of times with deleting the app first.

Climbatize: Ohhhhhh my bad !!! Friday review... I read a force unroll instead of a boolean manipulation 🤦‍♂️

Climbatize (approved):

PR #358, opened 2021-11-26
[MINI-4414] Default locale from host app

laubengaier:
# Description
Adding default locale from the host app so the host will control which locale should be used for the miniapp

- Starting small refactorings from delegates to closures (works with overriding)
- Locale is inferred through Localization.strings file which uses system locale in the background to fallback to the default device language
- Locale is validated by `{ISO 639-1}-{ISO 3166}` which is eg `ja-JP`

## Links
https://jira.rakuten-it.com/jira/browse/MINI-4414

# Checklist


Climbatize (approved):
LGTM

MiniApp/Classes/core/Protocols/MiniAppMessageDelegate.swift
```diff
@@ -95,10 +106,23 @@ public class MAHostEnvironmentInfo: Codable {
     let platformVersion: String
     let hostVersion: String
     let sdkVersion: String
+    let hostLocale: String
 
-    public init(platformVersion: String, hostVersion: String, sdkVersion: String) {
+    public init?(platformVersion: String, hostVersion: String, sdkVersion: String, hostLocale: String) {
+        guard hostLocale.isValidLocale else { return nil }
```
rleojoseph, line 112: I don't think we need to return nil here. Is that a requirement?  We should still return the `sdkVersion`, `hostVersion` & `platformVersion` even if `hostLocale` is not found right?

MiniApp/Classes/core/JavascriptBridge/MiniAppScriptMessageHandler.swift
```diff
@@ -87,7 +87,11 @@ internal class MiniAppScriptMessageHandler: NSObject, WKScriptMessageHandler {
         case .getPoints:
             fetchPoints(with: callbackId)
         case .getHostEnvironmentInfo:
-            getHostEnvironmentInfo(with: callbackId)
+            if hostAppMessageDelegate?.getEnvironmentInfo != nil {
+                getEnvironmentInfo(with: callbackId)
+            } else {
+                getHostEnvironmentInfo(with: callbackId)
+            }
```
rleojoseph, line 94: This else block will never get called right? 
`getHostEnvironmentInfo` is deprecated also now in this PR we have extension for `getEnvironmentInfo` ?

rleojoseph (changes requested):
Looks ok, just a few comments. Please check it. Also, could you please update Readme & Changelog as well ?

MiniApp/Classes/core/JavascriptBridge/MiniAppScriptMessageHandler.swift
laubengaier: @rleojoseph I was unsure if we should keep the support for the old delegate method because it might have been implemented so shouldn't the deprecated implementation work until we remove it with the release after?

MiniApp/Classes/core/Protocols/MiniAppMessageDelegate.swift
laubengaier: Also bit unsure here, I think there should not be a case where the hostLocale is nil because it will fallback to the default language if not set if we would not return nil we would have to think of how to handle the locale fallback for worst case. any suggestions?

rleojoseph: if `var getEnvironmentInfo: (() -> (MAHostEnvironmentInfo?))? { }` is overridden in Host app and  `miniapp.sdk.ios.locale` is not added in the Host app Localizable, then it will be nil.
If the method is not overridden then it won't be nil because SDK provides one.
Either you can remove this condition OR you can pass an empty string for `hostLocale` value OR send the default value from the SDK like below,

```
if !hostLocale.isValidLocale {
    self.hostLocale = MASDKLocale.localize(<LOCALE_KEY>)
}
```
What do you think ?

laubengaier: added your suggestion 👍 👍

MiniApp/Classes/core/JavascriptBridge/MiniAppScriptMessageHandler.swift
laubengaier: should be good now 💥

rleojoseph (approved):
Looks good to me :+1:

Only the test cases are failing. Please fix it before merging.

PR #361, opened 2021-12-06
[MINI-4419] Promotional image and text

laubengaier:
# Description
- Extending MiniAppInfo by `promotionalImageUrl` and `promotionalText`
- Adding Sharing VC to display the promotional content via `MiniAppUI`

https://user-images.githubusercontent.com/17570451/144808495-7b19d6b0-4064-47e7-ad6c-0f0b8488db36.MP4

## Links
https://jira.rakuten-it.com/jira/browse/MINI-4419

# Checklist

### DIFF
--- a/MiniApp/Classes/ui/MiniAppSharePreviewViewController.swift
+++ b/MiniApp/Classes/ui/MiniAppSharePreviewViewController.swift
@@ -40,9 +40,9 @@ public class MiniAppSharePreviewViewController: UIViewController {
         self.promotionImageUrl = promotionImageUrl
         super.init(nibName: nil, bundle: nil)
     }
-    
+
     required init?(coder: NSCoder) { return nil }
-    
+
     public override func viewDidLoad() {
         super.viewDidLoad()
 
@@ -52,18 +52,18 @@ public class MiniAppSharePreviewViewController: UIViewController {
 
         setupUI()
     }
-    
+
     public override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
         loadPreviewImage()
     }
-    
+
     func setupUI() {
         view.addSubview(promotionLabel)
         NSLayoutConstraint.activate([
             promotionLabel.topAnchor.constraint(equalTo: view.safeAreaLayoutGuide.topAnchor, constant: 20),
             promotionLabel.leadingAnchor.constraint(equalTo: view.leadingAnchor),
-            promotionLabel.trailingAnchor.constraint(equalTo: view.trailingAnchor),
+            promotionLabel.trailingAnchor.constraint(equalTo: view.trailingAnchor)
         ])
 
         view.addSubview(promotionImageView)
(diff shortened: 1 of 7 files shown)


MiniApp/Classes/ui/MiniAppViewController.swift
```diff
@@ -224,6 +234,37 @@ public class MiniAppViewController: UIViewController {
...
+                case .success(let info):
+                    guard
+                        let text = info.promotionalText,
+                        let imageUrl = info.promotionalImageUrl
+                    else {
+                        return
+                    }
+                    DispatchQueue.main.async {
+                        self.showShareAlert(text: text, imageUrl: imageUrl)
+                    }
```
Climbatize, line 254: Do you think we should capture self there?

```suggestion
                    DispatchQueue.main.async { [weak self] in
                        self?.showShareAlert(text: text, imageUrl: imageUrl)
                    }
```

MiniApp/Classes/ui/MiniAppSharePreviewViewController.swift
```diff
@@ -0,0 +1,107 @@
...
+            promotionImageView.leadingAnchor.constraint(equalTo: view.leadingAnchor, constant: 20),
+            promotionImageView.trailingAnchor.constraint(equalTo: view.trailingAnchor, constant: -20),
+            promotionImageView.heightAnchor.constraint(equalTo: view.widthAnchor, multiplier: 0.667, constant: -40)
+        ])
+
+        promotionImageView.addSubview(loadingIndicatorView)
+        NSLayoutConstraint.activate([
+            loadingIndicatorView.centerXAnchor.constraint(equalTo: promotionImageView.centerXAnchor),
+            loadingIndicatorView.centerYAnchor.constraint(equalTo: promotionImageView.centerYAnchor)
+        ])
```
Climbatize, line 81: All of this (and all the lazy variables declarations) could be simplified by the use of storyboards 😅

rleojoseph (approved):
Looks good to me :+1: Could you please update the following,

* Changelog
* Testcases (You can actually add MiniAppSharePreviewController in the ignore list as it has mostly UI code )

Just a few suggestions,
* Should we need to replace the Close button with X like button. I see Share image button and  Close text button doesn't go well together.

MiniApp/Classes/ui/MiniAppViewController.swift
laubengaier: I think dispatch.main should not cause much issues because it should be always called but for the outer closure it's probably safer 👍

MiniApp/Classes/ui/MiniAppSharePreviewViewController.swift
laubengaier: true but I prefer coded ui 😄  if required I can change it

laubengaier:
> Looks good to me 👍
> 
> @laubengaier Could you please update the following,
> 
> * Changelog
> * Testcases (You can actually add MiniAppSharePreviewController in the ignore list as it has mostly UI code )
> 
> Just a few suggestions,
> 
> * Should we need to replace the Close button with X like button. I see Share image button and  Close text button doesn't go well together.

@rleojoseph I've added the changelog. for test cases I think they are already in the ignore list (codecov.yml) -> 
```
ignore:
   - "Pods"
   - "Tests"
   - "Example"
   - "MiniApp/Classes/ui" <<<
   ...
```
I've also added the default (x) button but the navigationbar color makes it a bit hard to see
IMG_82C50E96A81B-1

corycaywood (approved):
Looks good 👍

MiniApp/Classes/ui/MiniAppSharePreviewViewController.swift
Climbatize: that's fine, one day we'll migrate to SwiftUI so we kind of have a taste of it

laubengaier: yeah best of both comes together then 🥇

PR #373, opened 2022-01-14
Fix for mini app memory leak

rleojoseph:
# Description
* Even after the mini-app is closed, due to strong references of these closures, `RealMiniAppView` object is not removed from memory. Added weak references for the closures so that it gets deleted when `RealMiniAppView` is removed 

# Checklist

### DIFF
--- a/Sources/Classes/core/Display/RealMiniAppView.swift
+++ b/Sources/Classes/core/Display/RealMiniAppView.swift
@@ -89,11 +89,11 @@ internal class RealMiniAppView: UIView {
     }
 
     fileprivate func initExternalWebViewClosures() {
-        onExternalWebviewResponse = { (url) in
-            self.webView.load(URLRequest(url: url))
+        onExternalWebviewResponse = { [weak self] (url) in
+            self?.webView.load(URLRequest(url: url))
         }
-        onExternalWebviewClose = { (url) in
-            self.didReceiveEvent(.externalWebViewClosed, message: url.absoluteString)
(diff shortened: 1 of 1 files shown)


laubengaier (approved):
lgtm 👍

Sources/Classes/core/Display/RealMiniAppView.swift
```diff
@@ -89,11 +89,11 @@ internal class RealMiniAppView: UIView {
...
     fileprivate func initExternalWebViewClosures() {
-        onExternalWebviewResponse = { (url) in
-            self.webView.load(URLRequest(url: url))
+        onExternalWebviewResponse = { [weak self] (url) in
+            self?.webView.load(URLRequest(url: url))
         }
-        onExternalWebviewClose = { (url) in
-            self.didReceiveEvent(.externalWebViewClosed, message: url.absoluteString)
+        onExternalWebviewClose = { [weak self] (url) in
+            self?.didReceiveEvent(.externalWebViewClosed, message: url.absoluteString)
```
laubengaier, line 96: did it capture `self` here as well? the others look more sus

rleojoseph: @laubengaier , yes, only when external webview is closed, however not mentioning this as weak reference wouldn't let the RealMiniAppView to call deinit method.

laubengaier: ok good to know 👍

Climbatize (approved):

PR #383, opened 2022-02-21
Retrieve mini-app directly from cache

rleojoseph:
# Description
* Retrieve mini-app from cache directly using `fromCache` variable.
  * Offline support is enable already, however SDK is trying to fetch the info of a mini-app when `create()` is called. This change would provide an option to the Host app to directly fetch the mini-app from cache.

## Links
[MINI-4777](https://jira.rakuten-it.com/jira/browse/MINI-4777)

# Checklist
- [ ] I have read the [contributing guidelines](CONTRIBUTING.md).
- [ ] I have completed the [SDK Development Learning Path](CONTRIBUTING.md#SDK-Development-Learning-Path) (if submitting a feature PR)

### DIFF
--- a/Sources/Classes/core/RealMiniApp.swift
+++ b/Sources/Classes/core/RealMiniApp.swift
@@ -1,3 +1,4 @@
+// swiftlint:disable file_length
 internal class RealMiniApp {
     var miniAppInfoFetcher: MiniAppInfoFetcher
     var metaDataDownloader: MetaDataDownloader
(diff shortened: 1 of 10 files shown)


Sources/Classes/core/RealMiniApp.swift
```diff
@@ -361,6 +386,35 @@ internal class RealMiniApp {
...
+                          queryParams: String? = nil,
+                          completionHandler: @escaping (Result<MiniAppDisplayDelegate, MASDKError>) -> Void,
+                          messageInterface: MiniAppMessageDelegate? = nil,
+                          adsDisplayer: MiniAppAdDisplayer? = nil) {
+        if appId.isEmpty {
+            return completionHandler(.failure(.invalidAppId))
+        }
+        if version.isEmpty {
+            return completionHandler(.failure(.invalidVersionId))
+        }
```
Climbatize, line 401: I'm still not sure we should test version number. @corycaywood, if we try to launch a new version that failed to be cached, should we return the latest version in cache?

Climbatize (commented):
Very good job. I still have a concern I commented. Let's discuss it internally

Sources/Classes/core/RealMiniApp.swift
corycaywood: Is this checking the version id which was passed in the `create` interface?
`versionId` is optional, right? So I think it should check the version id ONLY if it is not `nil`.

Otherwise, if version id is `nil`, then it should return whatever version that is cached.

rleojoseph: @corycaywood @Climbatize . Understood 👍🏻 . Updated-based feedback.

corycaywood (approved):
👍

Climbatize (approved):
LGTM

PR #386, opened 2022-03-15
[MINI-4672] Download Files to device

laubengaier:
# Description
New implementation for downloading files to device passing a url and filename instead of xhr (and base64)

## Links
https://jira.rakuten-it.com/jira/browse/MINI-4672

# Checklist


corycaywood (approved):
👍

rleojoseph (approved):
Looks good to me :+1:

Climbatize (approved):
LGTM 👍

Climbatize (commented):
Please document and add a changelog line before merging

Climbatize (changes requested):
I noticed that unit tests failed because of your changes, please ensure that your changes are retro-compatible with 4.0.0, put deprecations if needed

Sources/Classes/core/Protocols/MiniAppMessageDelegate.swift
```diff
@@ -89,3 +96,7 @@ public class MAHostEnvironmentInfo: Codable {
         )
     }
 }
+
+public class DownloadHeaders: Codable {
+    let token: String?
+}
```
corycaywood, line 102: @laubengaier Sorry, I missed seeing this before. The headers needs to be open ended and allow any number of generic String pair of headers, not just a single param.

Sources/Classes/core/JavascriptBridge/MiniAppScriptMessageHandler.swift
```diff
@@ -519,6 +521,24 @@ internal class MiniAppScriptMessageHandler: NSObject, WKScriptMessageHandler {
...
+            let fileName = parameters?.filename,
+            let url = parameters?.url,
+            let headers = parameters?.headers
+        {
+             hostAppMessageDelegate?.downloadFile(fileName: fileName, url: url, headers: headers) { (result) in
+                switch result {
+                case .success:
+                    self.executeJavaScriptCallback(responseStatus: .onSuccess, messageId: callbackId, response: fileName)
+                case .failure:
+                    self.executeJavaScriptCallback(responseStatus: .onError, messageId: callbackId, response: getMiniAppErrorMessage(MiniAppJavaScriptError.unexpectedMessageFormat))
```
corycaywood, line 536: I think we need to use a specific format for the Error, similar to what we did with `getAccessToken`. Also in the case that the download failed, we should pass the status code and message. This is what I suggested, but you can sync up with Android side:
```
{
  status: number,
  reason: String
}
```

corycaywood (changes requested):

Sources/Classes/core/Protocols/MiniAppMessageDelegate.swift
laubengaier: ok gotcha, replaced it with [String: String]

Sources/Classes/core/JavascriptBridge/MiniAppScriptMessageHandler.swift
laubengaier: added a new error `MASDKDownloadFileError` that should cover the error cases

laubengaier:
> I noticed that unit tests failed because of your changes, please ensure that your changes are retro-compatible with 4.0.0, put deprecations if needed

what do you specifically mean 😄 ? it's an addition before we could just intercept the tapped url

Climbatize (approved):
LGTM

corycaywood (approved):
Looks good 👍

PR #393, opened 2022-04-06
[MINI-1732] - In App Purchase implementation

rleojoseph:
# Description
* Updated SDK to handle the Purchase Product using product Id flow
* Added a new subspec for IAP
* Updated Demo app to handle IAP flow

## Links
[MINI-1732](https://jira.rakuten-it.com/jira/browse/MINI-1732)

# Checklist


Sources/Classes/in-app-purchase/MiniAppIAPModule.swift
```diff
@@ -0,0 +1,76 @@
...
+            case .purchasing, .deferred, .restored:
+                break
+            case .purchased:
+                SKPaymentQueue.default().finishTransaction($0)
+                sendTranactionDetails(transaction: $0, status: .purchased)
+            case .failed:
+                SKPaymentQueue.default().finishTransaction($0)
+                sendTranactionDetails(transaction: $0, status: .failed)
+            @unknown default:
+                print("Default", $0.payment.productIdentifier)
```
laubengaier, line 50: The print here can be replaced by the MiniAppLogger & sth more descriptive like `MiniAppLogger.d("MiniApp Transaction Update: Default \($0.payment.productIdentifier)")`

Example/Controllers/ViewController.swift
```diff
@@ -46,6 +48,11 @@ class ViewController: RATViewControllerWithTableView {
         self.tableView.refreshControl = refreshControl
         locationManager.delegate = self
         self.pageName = MASDKLocale.localize("demo.app.rat.page.name.home")
+        miniAppIAPModule = MiniAppIAPModule()
+        guard let inAppPurchaseModule = miniAppIAPModule else {
+            return
+        }
+        SKPaymentQueue.default().add(inAppPurchaseModule)
```
laubengaier, line 55: I think because the `miniAppModule` will be set anyways this could be written with an if which would be a bit more readable like below wdyt?

```suggestion
        if miniAppIAPModule == nil {
            let module = MiniAppIAPModule()
            self.miniAppIAPModule = module
            SKPaymentQueue.default().add(module)
        }
```

Sources/Classes/core/JavascriptBridge/MiniAppScriptMessageHandler.swift
```diff
@@ -52,7 +52,7 @@ internal class MiniAppScriptMessageHandler: NSObject, WKScriptMessageHandler {
         handleActionCommand(action: requestAction, requestParam: requestParam, callbackId: callbackId)
     }
 
-    func handleActionCommand(action: MiniAppJSActionCommand, requestParam: RequestParameters?, callbackId: String) {
+    func handleActionCommand(action: MiniAppJSActionCommand, requestParam: RequestParameters?, callbackId: String) { // swiftlint:disable:this function_body_length
```
laubengaier, line 55: this looks like the swiftlint statement is inside the function body but shouldn't it be on top of the funtion like so?

```suggestion
    // swiftlint:disable:this function_body_length
    func handleActionCommand(action: MiniAppJSActionCommand, requestParam: RequestParameters?, callbackId: String) {
```

Sources/Classes/in-app-purchase/MiniAppIAPModule.swift
```diff
@@ -0,0 +1,76 @@
...
+    public typealias ProductResponseCompletionHandler = ((Result<MAProductResponse, MAProductResponseError>)) -> Void
+    var productResponseHandlerObj: ProductResponseCompletionHandler?
+    var paymentProductObj: SKProduct?
+    private var models = [SKProduct]()
+    var productId: String = ""
+
+    public func initiateProductPayment(productId: String, completionHandler: @escaping ProductResponseCompletionHandler) {
+        self.productResponseHandlerObj = completionHandler
+        self.productId = productId
+        let productRequest = SKProductsRequest(productIdentifiers: Set(["com.temporary.id"]))
```
laubengaier, line 15: Just a question: the product identifiers are hard coded here is this intended for the current state?

laubengaier (commented):
looks good 👍 just a couple of questions

Sources/Classes/in-app-purchase/MiniAppIAPModule.swift
rleojoseph: Done

Example/Controllers/ViewController.swift
rleojoseph: Yep. Looks good, updated code now

Sources/Classes/core/JavascriptBridge/MiniAppScriptMessageHandler.swift
rleojoseph: Yeah, that's how we used to get rid of the warning right? But this one doesn't seem to work that way, when I added that. Now after cleaning up and added the disable command like you said. Its working fine now 🙄

Sources/Classes/in-app-purchase/MiniAppIAPModule.swift
rleojoseph: Sorry, this was meant to be cleaned up while working on testcases. Thanks for noticing. Removed it now

Sources/Classes/core/JavascriptBridge/MiniAppScriptMessageHandler.swift
laubengaier: 😄 the almighty clean & run

laubengaier (approved):

Sources/Classes/core/RealMiniApp.swift
```diff
@@ -438,6 +438,12 @@ internal class RealMiniApp {
 }
 
 extension RealMiniApp: MiniAppMessageDelegate {
+    #if RMA_SDK_IAP
```
rleojoseph, line 441: @Climbatize @corycaywood We didn't move the `MiniAppPurchaseDelegate` into a separate module/subspec.
Subspec contains only the Host app implementation, so I guess this check is not required.
Do you think we should move the MiniAppPurchaseDelegate as well into separate subspec ?

Climbatize: @rleojoseph I'm fine with you removing this check, as long as the Sample App compiles when you remove the IAP dependency from the podfile 👍

corycaywood: @rleojoseph I think it's correct to have `MiniAppPurchaseDelegate` in the main module, but like mentioned by @Climbatize it doesn't seem like this check is needed.

On Android, the main purposed of having a separate module is to separate the dependency on the Billing library. But in the case of iOS, it's still good to have a separate module just to separate out the implementation in case the Host App wants to use their own.

Sources/Classes/in-app-purchase/MiniAppIAPModule.swift
```diff
@@ -0,0 +1,76 @@
...
+            case .purchasing, .deferred, .restored:
+                break
+            case .purchased:
+                SKPaymentQueue.default().finishTransaction($0)
+                sendTranactionDetails(transaction: $0, status: .purchased)
+            case .failed:
+                SKPaymentQueue.default().finishTransaction($0)
+                sendTranactionDetails(transaction: $0, status: .failed)
+            @unknown default:
+                MiniAppLogger.d("MiniApp Transaction Update: Default \($0.payment.productIdentifier)")
```
corycaywood, line 50: What is the caae where this would get triggered? Should it send an error response to the mini app?

Sources/Classes/in-app-purchase/MiniAppIAPModule.swift
```diff
@@ -0,0 +1,76 @@
...
+            }
+        })
+    }
+
+    func sendTranactionDetails(transaction: SKPaymentTransaction, status: MAProductResponseStatus) {
```
corycaywood, line 55:
```suggestion
    func sendTransactionDetails(transaction: SKPaymentTransaction, status: MAProductResponseStatus) {
```

Sources/Classes/core/Models/InAppPurchaseResponse.swift
```diff
@@ -0,0 +1,73 @@
...
+        case .failedToConformToProtocol:
+        return "Host app failed to implement required interface"
+        }
+    }
+}
+
+public enum MAProductResponseStatus: String, Codable {
+    case purchased = "PURCHASED"
+    case failed = "FAILED"
+    case restored = "RESTORED"
```
corycaywood, line 26: I think the type in the JS SDK didn't have a field for the status? So how is this getting communicated to the Mini App?

corycaywood (commented):
Looking good, but I left a couple comments below.

Sources/Classes/in-app-purchase/MiniAppIAPModule.swift
rleojoseph: @corycaywood , we pretty much handle all the cases here from [TransactionState](https://developer.apple.com/documentation/storekit/skpaymenttransactionstate), however, Apple throws warning like this,

```
Switch covers known cases, but 'SKPaymentTransactionState' may have additional unknown values, possibly added in future versions
Handle unknown values using "@unknown default"
```

So added this default case with logs

Sources/Classes/core/Models/InAppPurchaseResponse.swift
rleojoseph: Yeah, updating JS SDK Side now. Thanks @corycaywood

Climbatize (approved):

Sources/Classes/core/JavascriptBridge/MiniAppScriptMessageHandler.swift
```diff
@@ -586,6 +589,23 @@ internal class MiniAppScriptMessageHandler: NSObject, WKScriptMessageHandler {
...
+            hostAppMessageDelegate?.purchaseProduct(withId: purchaseItemId, completionHandler: { result in
+                switch result {
+                case .success(let response):
+                    guard let encodedResponse = ResponseEncoder.encode(data: response) else {
+                        self.executeJavaScriptCallback(responseStatus: .onError, messageId: callbackId, response: getMiniAppErrorMessage(MiniAppJavaScriptError.internalError))
+                        return
+                    }
+                    self.executeJavaScriptCallback(responseStatus: .onSuccess, messageId: callbackId, response: encodedResponse)
+                case .failure(let error):
+                    self.executeJavaScriptCallback(responseStatus: .onError, messageId: callbackId, response: getMiniAppErrorMessage(error))
```
corycaywood, line 603: Maybe we should return an error object instead of just a String? Like we did with the access token and some other features.

I'm just thinking in the future there could be many different error cases that need to be handled. In general, it seems better to switch to using error objects instead of strings for all new features.

rleojoseph: Done

corycaywood (approved):

PR #397, opened 2022-04-26
[MINI-4968] Secure Storage Implementation

laubengaier:
# Description
Secure Storage Implementation

1. Added handling for new `MiniAppSecureStorage` feature to be basically handled in `MiniAppRealView`
2. Events sent from the Bridge will be send to the new `MiniAppSecureStorageDelegate`
3. SecureStorage is initialized in `MiniAppRealView` so it's loaded when the view will be loaded and unloaded when the View "deinits"

- [x] Unit Tests
- [x] Max Storage Space (Definition)
- [ ] JS SDK PR ready (https://github.com/rakutentech/js-miniapp/pull/218)

## Links
https://jira.rakuten-it.com/jira/browse/MINI-4968

# Checklist


Sources/Classes/core/JavascriptBridge/MiniAppScriptMessageHandler.swift
```diff
@@ -586,6 +605,119 @@ internal class MiniAppScriptMessageHandler: NSObject, WKScriptMessageHandler {
...
+        if let key = parameters?.secureStorageKey {
+            do {
+                if let value = try secureStorageDelegate?.get(key: key) {
+                    self.executeJavaScriptCallback(
+                        responseStatus: .onSuccess,
+                        messageId: callbackId,
+                        response: value
+                    )
+                } else {
+                    self.executeJavaScriptCallback(responseStatus: .onError, messageId: callbackId, response: prepareMAJavascriptError(MiniAppSecureStorageError.storageKeyNotFound))
```
corycaywood, line 618: I think we should just return a `null` response in case the key doesn't exist.

@laubengaier @leorjoseph @munir-rakuten What do you think?

Sources/Classes/core/Storage/MiniAppSecureStorage.swift
```diff
@@ -0,0 +1,256 @@
...
+        case .unknown:
+            return "unknown"
+        case .storageNotExistent:
+            return "storageNotExistent"
+        case .storageFileEmpty:
+            return "storageFileEmpty"
+        case .storageBusyProcessing:
+            return "storageBusyProcessing"
+        case .storageKeyNotFound:
+            return "storageKeyNotFound"
```
corycaywood, line 238: The `name` needs to match what's defined on the JS side so that it can be parsed to the correct error type (but I don't think @leorjoseph has added the error types yet in JS).

The names I suggested are `SecureStorageFullError` and `SecureStorageIOError` but maybe we also need some others.

Sources/Classes/core/Storage/MiniAppSecureStorage.swift
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    public static func clearSecureStorage(for miniAppId: String) throws {
+        MiniAppLogger.d("🔑 Secure Storage: destroy")
+        try FileManager.default.removeItem(at: storagePath(appId: miniAppId))
+    }
+
+    public func size() throws -> UInt64 {
+        let fileSize = MiniAppSecureStorage.storagePath(appId: appId).fileSize
+        guard fileSize > 0 else { throw MiniAppSecureStorageError.storageFileEmpty }
```
corycaywood, line 202: Would this be the case that the Secure Storage failed to load/hasn't finished loading yet?
Or is this the case where the user hasn't set any items yet?

I think the case that the user hasn't set any items, we don't need to return an error and we can just return 0.
And in the case that Secure Storage hasn't loaded, we can return some specific error like `SecureStorageNotLoaded`.

What do you think?

corycaywood (commented):
Looking good - I left a few comments below.

Sources/Classes/core/Storage/MiniAppSecureStorage.swift
laubengaier: @corycaywood the minimum size for the plist is around 42 bytes so it should always reflect if there is no store file yet (plist is in xml format so apple adds a few bits even if it's empty). So `0` is no file and `> 0` we've got a store. But sure I can remove the error check and just return the size no matter what the store doesn't need to be loaded for the size method. wdyt?

Sources/Classes/core/Storage/MiniAppSecureStorage.swift
laubengaier: ok sounds good I'll wait for the specs

Sources/Classes/core/JavascriptBridge/MiniAppScriptMessageHandler.swift
laubengaier: sure makes sense, I'll update it

Sources/Classes/core/Display/RealMiniAppView.swift
```diff
@@ -138,6 +145,14 @@ internal class RealMiniAppView: UIView {
...
         initExternalWebViewClosures()
         observeWebView()
+
+        secureStorage.loadStorage { success in
+            if success {
+                NotificationCenter.default.sendCustomEvent(MiniAppEvent.Event(type: .secureStorageReady, comment: "MiniApp Secure Storage Ready"))
+            } else {
+                NotificationCenter.default.sendCustomEvent(MiniAppEvent.Event(type: .secureStorageError, comment: "MiniApp Secure Storage Error"))
+            }
+        }
```
rleojoseph, line 155: Can we move this into seperate method

Sources/Classes/core/Display/RealMiniAppView.swift
```diff
@@ -242,6 +257,7 @@ internal class RealMiniAppView: UIView {
         UIViewController.attemptRotationToDeviceOrientation()
         webView.configuration.userContentController.removeMessageHandler()
         NotificationCenter.default.removeObserver(self)
+        secureStorage.unloadStorage()
```
rleojoseph, line 260: We have to make sure we test this thing that we don't have any memory leak.

rleojoseph (commented):
Overall looks ok. Just a few comments. Could you please update Changelog and documentation.

Sources/Classes/core/Display/RealMiniAppView.swift
laubengaier: @rleojoseph deinit is called for both I'll add a deinit logging for MiniAppSecureStorage. I think for more proper Unit Tests we need a byte more time.

corycaywood (approved):
Looks good 👍  I think just need to update the error types.

rleojoseph (approved):
Looks good to me :+1:

PR #400, opened 2022-05-09
Replace MessageDelegate `getSecureStorageSizeLimit` with MiniAppSDKConfig property

laubengaier:
# Description
- replace message delegate maxStorage delegate with MiniAppSdkConfig.storageMaxSizeInBytes
- check every time a value is set for exceeding size

# Checklist

### DIFF
--- a/Sources/Classes/core/RealMiniApp.swift
+++ b/Sources/Classes/core/RealMiniApp.swift
@@ -20,29 +20,29 @@ internal class RealMiniApp {
         self.init(with: nil)
     }
 
-    init(with settings: MiniAppSdkConfig?, and navigationSettings: MiniAppNavigationConfig? = nil, sslPinningSettings: MiniAppSSLConfig? = nil) {
+    init(with config: MiniAppSdkConfig?, and navigationConfig: MiniAppNavigationConfig? = nil, sslPinningSettings: MiniAppSSLConfig? = nil) {
         miniAppInfoFetcher = MiniAppInfoFetcher()
         metaDataDownloader = MetaDataDownloader()
-        miniAppClient = MiniAppClient(baseUrl: settings?.baseUrl,
-                                      rasProjectId: settings?.rasProjectId,
-                                      subscriptionKey: settings?.subscriptionKey,
-                                      hostAppVersion: settings?.hostAppVersion,
-                                      isPreviewMode: settings?.isPreviewMode)
+        miniAppClient = MiniAppClient(baseUrl: config?.baseUrl,
+                                      rasProjectId: config?.rasProjectId,
+                                      subscriptionKey: config?.subscriptionKey,
+                                      hostAppVersion: config?.hostAppVersion,
+                                      isPreviewMode: config?.isPreviewMode)
         manifestDownloader = ManifestDownloader()
         miniAppStatus = MiniAppStatus()
         miniAppPermissionStorage = MiniAppPermissionsStorage()
         miniAppManifestStorage = MAManifestStorage()
         miniAppDownloader = MiniAppDownloader(apiClient: miniAppClient, manifestDownloader: manifestDownloader, status: miniAppStatus)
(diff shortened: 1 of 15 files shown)


Sources/Classes/core/RealMiniApp.swift
```diff
@@ -33,13 +33,14 @@ internal class RealMiniApp {
         miniAppPermissionStorage = MiniAppPermissionsStorage()
         miniAppManifestStorage = MAManifestStorage()
         miniAppDownloader = MiniAppDownloader(apiClient: miniAppClient, manifestDownloader: manifestDownloader, status: miniAppStatus)
-        displayer = Displayer(navigationSettings)
+        displayer = Displayer(settings, navigationSettings)
```
rleojoseph, line 36: Can we rename this to `config` instead of `settings` ? And also the places where we have used. Because so far in SDK, we have shared this object in different classes with `config` naming

Sources/Classes/core/Storage/MiniAppSecureStorage.swift
```diff
@@ -32,8 +32,9 @@ public class MiniAppSecureStorage: MiniAppSecureStorageDelegate {
     private static let storageName: String = "securestorage"
     static var storageFullName: String { return storageName + ".plist" }
 
-    public init(appId: String) {
+    public init(appId: String, storageMaxSizeInBytes: UInt64? = nil) {
         self.appId = appId
+        self.fileSizeLimit = storageMaxSizeInBytes ?? 2_000_000
```
rleojoseph, line 37: We should update this in documentation as well. Default size is supported by the SDK itself.
@corycaywood , I think we should update the same in storage documentation?

Sources/Classes/core/Storage/MiniAppSecureStorage.swift
```diff
@@ -90,29 +91,53 @@ public class MiniAppSecureStorage: MiniAppSecureStorageDelegate {
...
+                let memorySize = try? getMemoryStorageFileSize()
+            else { completion?(.failure(.storageIOError))
+                return
+            }
+            let estimatedAddSize = UInt64(estimateAddData.count)
+            let estimatedFinalSize = memorySize + estimatedAddSize
+            guard estimatedFinalSize <= fileSizeLimit else {
+                completion?(.failure(.storageFullError))
+                return
+            }
```
rleojoseph, line 127: So we are making sure that the limit is not exceeded before making any new entry into the storage file. 🤔 
Can we move this to a seperate method?

rleojoseph (commented):
Overall looks ok to me. Just a few comments. Please check

Also, could you also update the Sample app code to send the filesize limit when creating a mini-app.

Sources/Classes/core/RealMiniApp.swift
laubengaier: @rleojoseph if it's ok I'll rename it for displayer to `sdkConfig` because we already had a parameter for the navigationConfig called `config`. so I'd change it to:

```
    init(_ sdkConfig: MiniAppSdkConfig? = nil, _ navConfig: MiniAppNavigationConfig? = nil) {
        self.settings = sdkConfig
        self.navConfig = navConfig
    }
```

Sources/Classes/core/Storage/MiniAppSecureStorage.swift
laubengaier: I've updated it with a slight improvement so I'm checking if the whole dict is sufficient to fit into the available storage instead for every key + extraction

Sources/Classes/core/RealMiniApp.swift
laubengaier: renamed it as well for `RealMiniApp` but kept it for `MiniApp` so we don't have to rename all `.shared` calls if that's ok

rleojoseph (approved):
Looks good to me :+1:

corycaywood (approved):

PR #404, opened 2022-05-19
[MINI-5030] Database for Secure Storage

laubengaier:
# Description
Secure Storage Database Implementation
- added adapter like pattern so the way secure storage is connected can be easily swapped out
- adding sqlite.swift as main secure storage database
- making sure all storage unit tests pass

## Links
https://jira.rakuten-it.com/jira/browse/MINI-5030

# Checklist


Sources/Classes/core/SecureStorage/Database/MiniAppSecureStorageSqliteDatabase+Model.swift
```diff
@@ -0,0 +1,46 @@
...
+        static func find(database: Connection, key: String) throws -> Entry? {
+            for entry in try database.prepare(entries.filter(Self.key == key)) {
+                return Entry(key: entry[Self.key], value: entry[Self.value])
+            }
+            return nil
+        }
+
+        static func upsert(database: Connection, key: String, value: String) throws {
+            let upsert = entries.upsert(Self.key <- key, Self.value <- value, onConflictOf: Self.key)
+            _ = try database.run(upsert)
```
Climbatize, line 32: can we return the results of upsert/delete etc. and log when something went wrong? would be cleaner in my opinion

Climbatize (commented):
Looks good, check my comments

corycaywood (approved):
Looks good 👍

Sources/Classes/core/SecureStorage/Database/MiniAppSecureStorageSqliteDatabase.swift
```diff
@@ -0,0 +1,112 @@
...
+        guard storageExists else {
+            completion?(.storageUnvailable)
+            return
+        }
+        do {
+            let dbQueue = try Connection(storagePath)
+            self.dbQueue = dbQueue
+            completion?(nil)
+        } catch {
+            print(error)
```
rleojoseph, line 60: May be MiniAppLogger.d

Sources/Classes/core/SecureStorage/Database/MiniAppSecureStorageDatabase.swift
```diff
@@ -0,0 +1,94 @@
+import Foundation
+
+public protocol MiniAppSecureStorageDatabase: AnyObject {
```
rleojoseph, line 3: Do we need to have this as `public` ?

Sources/Classes/core/SecureStorage/MiniAppSecureStorage.swift
```diff
@@ -0,0 +1,177 @@
...
+    var fileSizeLimit: UInt64
+
+    /// state if the storage is currently loading
+    var isStoreLoading: Bool = false
+
+    static var storageName: String { return "securestorage" }
+
+    let database: MiniAppSecureStorageDatabase
+
+    public init(
```
rleojoseph, line 19: Do we need to have this as `public` ?

rleojoseph (commented):
You have done a great job. Thanks man. Looks like 3rd party lib is inevitable. Overall looks ok to me. Just few comments.

Sources/Classes/core/Display/RealMiniAppView.swift
```diff
@@ -318,11 +318,11 @@ internal class RealMiniAppView: UIView {
     }
 
     func wipeAllSecureStorageForMiniApps() {
-        MiniAppSecureStorage.wipeSecureStorages()
+        try? MiniAppSecureStorage.wipeSecureStorages()
```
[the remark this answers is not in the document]
laubengaier: not crash but throw an error when eg the miniapp folder does not exists

Sources/Classes/core/SecureStorage/Database/MiniAppSecureStorageSqliteDatabase+Model.swift
laubengaier: done 👍

Sources/Classes/core/Display/RealMiniAppView.swift
```diff
@@ -318,11 +318,11 @@ internal class RealMiniAppView: UIView {
...
     }
 
     func clearSecureStorage(for miniAppId: String) {
```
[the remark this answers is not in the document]
laubengaier: yeah true I'll remove them

Sources/Classes/core/SecureStorage/Database/MiniAppSecureStorageDatabase.swift
```diff
@@ -0,0 +1,94 @@
...
+                }
+            } catch let error {
+                MiniAppLogger.d("🔑 Secure Storage Wipe Failed: \(name)", error.localizedDescription)
+                throw MiniAppSecureStorageError.storageIOError
+            }
+        }
+    }
+
+    static func wipe(for miniAppId: String) throws {
+        if !miniAppId.isEmpty {
```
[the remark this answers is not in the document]
laubengaier: hmm could be a case where the miniAppId is empty but I think only an edge case.. I don't mind

Sources/Classes/core/SecureStorage/Database/MiniAppSecureStorageSqliteDatabase.swift
```diff
@@ -0,0 +1,112 @@
...
+        dbQueue = nil
+    }
+
+    func find(for key: String) throws -> Entry? {
+        guard let dbQueue = dbQueue else { throw MiniAppSecureStorageError.storageUnvailable }
+        return try Entry.find(database: dbQueue, key: key)
+    }
+
+    func save(completion: ((Swift.Result<Bool, MiniAppSecureStorageError>) -> Void)? = nil) throws {
+        completion?(.success(true))
```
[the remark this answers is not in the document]
laubengaier: it's only necessary when the adapter changes, for the plist (database) adapter it was necessary to execute save. if we want to work in transactions or anything we might have to use it again. for sqlite it's currently not necessary so just a return true

Sources/Classes/core/SecureStorage/MiniAppSecureStorage.swift
laubengaier: same as above

Sources/Classes/core/SecureStorage/Database/MiniAppSecureStorageSqliteDatabase.swift
```diff
@@ -0,0 +1,112 @@
...
+
+    var storageFullName: String { return Self.storageFullName }
+
+    var storagePath: String {
+        let databasePath = "/\(appId)/\(MiniAppSecureStorageSqliteDatabase.storageFullName)"
+        let databaseUrl = FileManager.getMiniAppFolderPath().appendingPathComponent(databasePath)
+        let databaseUrlPath = databaseUrl.path
+        return databaseUrlPath
+    }
+    var storageExists: Bool {
```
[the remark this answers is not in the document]
laubengaier: renamed it to `doesStoragePathExist`

Sources/Classes/core/SecureStorage/Database/MiniAppSecureStorageDatabase.swift
rleojoseph: In LINK? Adding test cases in our SDK isn't enough? Just curios.
We need to limit exposing the Public interfaces/models unless its really needed.

Sources/Classes/core/SecureStorage/Database/MiniAppSecureStorageDatabase.swift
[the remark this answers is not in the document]
rleojoseph: Yeah, that's a edge case. Its fine, I think

rleojoseph (approved):
Overall Looks good to me :+1:

Please take a look at `public` interface comment. Also, could you please add Changelog and Readme file as well.

Sources/Classes/core/SecureStorage/Database/MiniAppSecureStorageDatabase.swift
laubengaier: @rleojoseph Yeah I mean we can go without making it public but that caused quite some headaches in the past because we need to fulfill the 80% coverage on Link as well and no possibility to override or mock things.. I was thinking about making a bridge for the MiniApp SDK in Link as well which we ignore from coverage that would solve some issues. wdyt?

rleojoseph: Yeah. That would be great. Instead of exposing the entire class/protocol we can have a bridge. We can even add these files in code cov ignore list if we had written testcases in SDK.

I guess you can add the changelog for the storage mechanism and for README, we need to add the documentation set of interfaces which will exposed to Host app

laubengaier: ok sounds good 👍  I'll remove the public modifiers and will prepare the bridge soon

rleojoseph (approved):
Looks good to me :+1:

PR #409, opened 2022-06-07
[MALI-1042] Save Image Activity Issue

laubengaier:
# Description
Saving an image results in the presented vc also to dismiss.
Replace UIActivityController with MiniAppActivityController which replaces the `dismiss` function

https://user-images.githubusercontent.com/17570451/172297520-747010de-12ba-457d-873c-c2c6010d6f96.MP4

## Links
https://jira.rakuten-it.com/jira/browse/MALI-1042

# Checklist

### DIFF
--- a/Example/Controllers/Extensions/ViewController+Activity.swift
+++ b/Example/Controllers/Extensions/ViewController+Activity.swift
@@ -1,0 +1,6 @@
+import Foundation
+import UIKit
+
+public class MiniAppActivityController: UIActivityViewController {
+    public override func dismiss(animated flag: Bool, completion: (() -> Void)? = nil) {}
+}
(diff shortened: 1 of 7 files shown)


Example/Controllers/Extensions/ViewController+Activity.swift
```diff
@@ -0,0 +1,6 @@
+import Foundation
+import UIKit
+
+public class MiniAppActivityController: UIActivityViewController {
+    public override func dismiss(animated flag: Bool, completion: (() -> Void)? = nil) {}
```
rleojoseph, line 5: So we don't need to call `super` here 🤔

rleojoseph (commented):
Also, could you please remove this https://github.com/rakutentech/ios-miniapp/blob/master/Sources/Classes/ui/MiniAppUI.swift#L115 extension?

Looks like you forgot to commit the Sample app project settings as well.

Also, could you please update Changelog about this fix, under sample app section

Example/Controllers/Extensions/ViewController+Activity.swift
laubengaier: that's the fix it doesn't dimiss anything but the activity vc (works like a joy)

rleojoseph (approved):
Looks good to me :+1:

PR #426, opened 2022-08-17
MINI-5307 Revamp MiniApp SDK

laubengaier:
# Description
MiniApp SDK Revamp
- Moving from `MiniApp.shared().create` to `MiniAppView`
- SwiftUI Support (UI Module Feature)
- Async/Await Support for `load()`

## Links
https://jira.rakuten-it.com/jira/browse/MINI-5307

# Checklist


Sources/Classes/core/View/MiniAppView.swift
```diff
@@ -0,0 +1,220 @@
...
+                }
+            }
+        }
+    }
+
+    public var alertInfo: CloseAlertInfo? {
+        return miniAppHandler.miniAppShouldClose()
+    }
+
+    public enum LoadStatus {
```
rleojoseph, line 217: May be mini-app load status?

rleojoseph (commented):
Definitely a great job @laubengaier . Overall it looks fine to me. I feel it would be great a peer review session because I have a lot of questions,

* Changes for Demo app looks great, but I don't think we need these many changes right now. Let's try to keep it simple so that the demo app is similar in both platform. Wireframes will be ready soon and we can align with the same.
* I don't understand the need for **MiniAppSUIView**, do you think we can move this to UI module?
* I see we are trying to expose too many interfaces to public, lets's talk about this in the discussion
* We need to add a lot of documentation for the public interfaces, it will be really difficult for the Host app integration.

Sources/Classes/core/Models/MiniAppManifest.swift
```diff
@@ -183,11 +183,13 @@ public struct MiniAppManifest: Codable, Equatable {
              versionId
     }
 
-    init(requiredPermissions: [MASDKCustomPermissionModel]?,
-         optionalPermissions: [MASDKCustomPermissionModel]?,
-         customMetaData: [String: String]?,
-         accessTokenPermissions: [MASDKAccessTokenScopes]?,
-         versionId: String?) {
+    public init(
```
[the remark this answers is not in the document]
laubengaier: Yeah the struct is already `public` and conforms to `Codable` so imho the initializer can be public too because it's accessible to same way by parsing a json into `MiniAppInfo`.
It makes it more convenient to initialize for tests and eg. one case in `MiniAppTermsView` to be able to preview different kinds of infos.
wdyt?

rleojoseph (approved):
Overall look good to me. Please do the following things before merging the PR,
* Please make sure the public interfaces are aligned with Android platform such as load, loadAsync
* Update the description for all public interfaces/class. We need more description for MiniAppView and MiniAppViewable especially
* Update the Changelog.md
* Update the Userguide.md file with the new changes and how this can be integrated (you can do this in a separate ticket if needed)

I see you moved the sample app change to a separate PR, thank you.

Sources/Classes/core/View/MiniAppView.swift
```diff
@@ -0,0 +1,218 @@
...
+
+    internal var stateImageView: UIImageView = {
+        let view = UIImageView()
+        view.translatesAutoresizingMaskIntoConstraints = false
+        return view
+    }()
+
+    internal var activityIndicatorView: UIActivityIndicatorView = {
+        let view = UIActivityIndicatorView(style: .medium)
+        view.color = .systemGray2
```
Climbatize, line 25: this color should be customizable and stick to the host app accent color

Sources/Classes/core/View/MiniAppView.swift
```diff
@@ -0,0 +1,218 @@
...
+        view.translatesAutoresizingMaskIntoConstraints = false
+        return view
+    }()
+
+    internal var activityLabel: UILabel = {
+        let view = UILabel()
+        view.translatesAutoresizingMaskIntoConstraints = false
+        view.text = "-"
+        view.textAlignment = .center
+        view.textColor = .systemGray
```
Climbatize, line 35: same

Climbatize (approved):
Very good PR, please take a look at my suggestions

PR #428, opened 2022-09-02
[MALI- 1161] - Fix Preview-codes backslash issue

rleojoseph:
# Description
QR Code feature is not working in PROD and testflight builds. Pretty strange that it happens only in Prod builds.
Removing any backslash and \\ in preview token that could break the preview feature in the PROD. We suspect that this could be the reason for the issue.
Preview tokens adds a extra / sometimes, this could help avoid the situation.

## Links
[MALI-1161](https://jira.rakuten-it.com/jira/browse/MALI-1161)

# Checklist

### DIFF
--- a/Sources/Classes/core/API/PreviewMiniappAPI.swift
+++ b/Sources/Classes/core/API/PreviewMiniappAPI.swift
@@ -16,7 +16,7 @@ internal class PreviewMiniappAPI {
 
     private func getMiniAppInfo(using token: String) -> URL? {
         let previewToken = token.replacingOccurrences(of: "\\\\/", with: "", options: .regularExpression, range: nil)
-        guard var baseURL = environment.baseUrl else {
+        guard let baseURL = environment.baseUrl else {
             return nil
         }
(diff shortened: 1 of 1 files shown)


Sources/Classes/core/API/PreviewMiniappAPI.swift
```diff
@@ -15,9 +15,10 @@ internal class PreviewMiniappAPI {
     }
 
     private func getMiniAppInfo(using token: String) -> URL? {
+        let previewToken = token.replacingOccurrences(of: "\\\\/", with: "", options: .regularExpression, range: nil)
```
Climbatize, line 18: Do you really need a regex for that, seems over-engineering to me ^^

Climbatize (approved):
LGTM, good finding!

Sources/Classes/core/API/PreviewMiniappAPI.swift
rleojoseph: Yeah, I was thinking of an easy approach but sometimes what if we get too many / in the preview code? in that case regex would be easy solution

laubengaier (approved):
lgtm!

PR #452, opened 2022-11-08
[No Ticket]Publishing AppCenter builds from CircleCI

rleojoseph:
# Description
* Changes will help us to upload the Builds to AppCenter

# Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,7 @@ container_medium: &container_medium
   resource_class: medium
 
 jobs:
-  build-sim:
+  build-and-test:
     <<: *container_medium
     environment:
       FL_OUTPUT_DIR: output
@@ -27,7 +27,12 @@ jobs:
       - run: pod install --repo-update
       - run:
           name: Build Simulator
-          command: fastlane build_sim_ci
+          command: fastlane build_simulator
+      - run:
+          name: Execute Tests
+          command: fastlane ci
+      - store_artifacts:
+          path: ./artifacts/unit-tests
       - persist_to_workspace:
             root: .
             paths:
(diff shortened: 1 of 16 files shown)


laubengaier (approved):
lgtm 🎉

.circleci/config.yml
```diff
@@ -43,6 +52,22 @@ jobs:
...
+      - run:
+          name: Copy Configurations
+          command: cp MiniApp-Secrets.xcconfig ../MiniApp-Secrets.xcconfig
+      - run:
+          name: Deploy to AppCenter
+          no_output_timeout: 20m
+          command: ./scripts/deploy-to-app-center.sh
+      - run:
+          name: Clear Cache
+          command: rm ../MiniApp-Secrets.xcconfig
```
Climbatize, line 69: what happens if deploy fails? is this command run anyway?

scripts/deploy-to-app-center.sh
```diff
@@ -35,10 +35,11 @@ appcenter distribute release \
 --app "$APP_CENTER_APP_NAME" \
 --release-notes-file "$TMP_DIR"/CHANGELOG.md \
 --group "$APP_CENTER_GROUP" \
---build-version "$APP_CENTER_BUILD_VERSION" \
+--build-version "$CIRCLE_BUILD_NUM" \
```
Climbatize, line 38: do we need to modify the script? this will make it tied to Circle CI

scripts/deploy-to-app-center.sh
```diff
@@ -117,6 +118,7 @@ appcenter distribute release \
 --file "$TMP_DIR"/MiniApp_Example.ipa \
 --quiet
```
Climbatize, line 119: if you want to log more info you can also remove the `--quiet`

Climbatize (commented):
Good one, please look at my concerns

.circleci/config.yml
rleojoseph: Right. I think we can skip this even if it is failed.

scripts/deploy-to-app-center.sh
rleojoseph: I was using `APP_CENTER_BUILD_VERSION`, but for some reasons, it wasn't showing in the Distribution groups. But now that we have a new group in AppCenter, I guess I can revert back this change

scripts/deploy-to-app-center.sh
rleojoseph: Nah. That would be too noisy. We can remove this whenever we need it.

Climbatize (approved):
I was mostly concerned by security if line 66 fails and then line 69 is not ran. It you ensured it's not the case then let's merge this PR :)

rleojoseph:
Let's hold this PR. I will make necessary changes

GondiTilak (approved):
Looks good to me :+1:

PR #475, opened 2023-01-24
[MINI-5708] Back and forward arrows show's blank white screen.

GondiTilak:
# Description
* Issue: Back and forward arrows show's blank white screen.

* Cause: iOS-SDK returns databaseunavailable error before creating the securestorage.sqlite file, when opening the MiniApps and it will never be able to find securestorage.sqlite file.

* Fix: in MiniAppSecureStorageSqliteDatabase.swift & MiniAppSecureStorage.swift files to remove the guard statement on load.

## Links
[MINI-5708](https://jira.rakuten-it.com/jira/browse/MINI-5708)
[MIINI-5708-fix demo](https://rak.box.com/s/01uulek2cbdc2g53t2bsuqfjk8y6kss3)

# Checklist

### DIFF
--- a/Sources/Classes/core/SecureStorage/Database/MiniAppSecureStorageSqliteDatabase.swift
+++ b/Sources/Classes/core/SecureStorage/Database/MiniAppSecureStorageSqliteDatabase.swift
@@ -63,10 +63,6 @@ class MiniAppSecureStorageSqliteDatabase: MiniAppSecureStorageDatabase {
     }
 
     func load(completion: ((MiniAppSecureStorageError?) -> Void)?) {
-        guard doesStoragePathExist else {
-            completion?(.storageUnavailable)
-            return
-        }
         do {
             let queue = try Connection(storagePath)
             self.dbQueue = queue
--- a/Sources/Classes/core/SecureStorage/MiniAppSecureStorage.swift
+++ b/Sources/Classes/core/SecureStorage/MiniAppSecureStorage.swift
@@ -33,20 +33,14 @@ class MiniAppSecureStorage: MiniAppSecureStorageDelegate {
     // MARK: - Load/Unload
     func loadStorage(completion: ((Bool) -> Void)? = nil) {
         isStoreLoading = true
-        DispatchQueue.global(qos: .userInitiated).async { [weak self] in
-            self?.database.load(completion: { [weak self] error in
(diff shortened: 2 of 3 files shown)


Sources/Classes/core/SecureStorage/Database/MiniAppSecureStorageSqliteDatabase.swift
```diff
@@ -63,10 +63,6 @@ class MiniAppSecureStorageSqliteDatabase: MiniAppSecureStorageDatabase {
     }
 
     func load(completion: ((MiniAppSecureStorageError?) -> Void)?) {
-        guard doesStoragePathExist else {
-            completion?(.storageUnavailable)
-            return
-        }
```
rleojoseph, line 69: Not sure why we want to remove this code? Could you please provide the reason why you want to remove it? It is expected that we will not DB file until we store some value in the DB, and it will return `storageUnavailable`

rleojoseph (changes requested):

Sources/Classes/core/SecureStorage/Database/MiniAppSecureStorageSqliteDatabase.swift
GondiTilak: Because, if we do not remove then on load storage call the result will be error always "storageUnavailable". so on loading the MiniApp the secure storage is not available all. Either we should stop sending the response to js-bridge on loading the mini app to avoid the behavior of blank screen and also handle the error processing In the js sample app. Then there will be no need of any code changes in the iOS SDK I believe.

rleojoseph: As we discussed, Secure storage ready notifications can't be removed. This seems to have an issue more on the JS Bridge or sample app. Please fix the same

GondiTilak: Got It, Thanks for the clarification. We will check the QA recommendation and proceed with the fix in JS-Sample app.
We are good to close this PR as these changes are not required on the iOS-SDK side.

GondiTilak:
Closing this PR and continuing with further investigation on the JS-Sample app.

